PR #5, opened 2019-05-30
Make module installable with Composer

dverkade:
**Description**

It's Magento best practice to install 3rd party vendor modules with Composer. This way a module can easily be updated to the latest version. Currently the Reach Payment plugin can't be installed with Composer. This PR addresses this issue and makes the module installable with Composer. Ideally Reach adds this to the packagist.org repository so that you won't have to add another repository to the root composer.json file in the Magento project.

**What has been done**

- Added composer.json file so that module will be installable with Composer
- Moved files to the main root according to Magento 2 best practices
- Added missing convertCurrency function which as introduced in commit: https://github.com/withreach/magento-plugin/commit/46a63d5a2d6107eae047abd43b422f54285df51f#diff-fa08262412d51d8e7d8e312c037cd915

**Next steps**

- Accept and process this Pull Request
- Add module on packagist.org so that the composer package can be installed without adding another repository in the root composer.json file of your Magento project.


dverkade:
Hi @lizzynibali ,

This PR is closed without accepting changes and the issues in the code are still present. Please reopen and validate PR to merge into the module.

PR #14, opened 2019-11-06
Mag 8

cstech:
Included is the new version of the plugin which is now at 1.0.3
This will trigger a database schema update to add country_of_origin to reach_hs_codes table.
Added the country_of_origin to the export file of hs codes


Magento 2.2 & 2.3/Reach/Payment/Setup/UpgradeSchema.php
```diff
@@ -282,6 +283,22 @@ public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $con
                 ]
             );
         }
+        
+        if (version_compare($context->getVersion(), "1.0.4") < 0) {
```
lizzyWithReach, line 287: maybe we can talk about this on our call tomorrow, the way I'm reading this is that if the version is less than 1.0.4 then the country of origin column will be added to the database schema. 

Would it be the case that if someone installed 1.0.3, which is less than 1.0.4, then upgraded later they could end up with more than one country of origin column in their database?

lizzyWithReach: so because if they have version 1.0.3  and less then it will execute on the update.  that’s why it’s <0 because it’s backwards compatible
version_compare(1.0.4, 1.0.4) will return 0
version_compare(1.0.5, 1.0.4)  returns 1
version_compare(1.0.3, 1.0.4)  returns -1
the database schema update checks or the existence of the column trying to be added
also the status would not change, you have to update the version in the  module.xml file in the PR
so until we change that version it will not try to execute the database schema.  it’s in the comments of the jira task.

lizzyWithReach (approved):

PR #20, opened 2019-11-08
Importing country of origin from csv

cstech:
Import CSV will now take in country of origin in the new column.
Country of origin will only take in two character codes and will uppercase them.  If the code is less or more than 2 characters the importer will not insert the country of origin


Reach/Payment/Model/ResourceModel/Carrier/CsvHsCode/CSV/RowParser.php
```diff
@@ -82,8 +85,25 @@ private function getHsCode(array $rowData, ColumnResolver $columnResolver)
...
+     * @param array $rowData
+     * @param ColumnResolver $columnResolver
+     * @return int|string
+     * @throws ColumnNotFoundException
+     */
+    private function getCountryOfOrigin(array $rowData, ColumnResolver $columnResolver)
+    {
+        $countryOfOrigin = $columnResolver->getColumnValue(ColumnResolver::COLUMN_COUNTRYOFORIGIN, $rowData);
+        if (strlen($countryOfOrigin) != 2) {
+            $countryOfOrigin = null;
```
lizzyWithReach, line 125: is there a way to return or log an error in this case? So that the merchant can know what happened at least?

Reach/Payment/Model/ResourceModel/Carrier/CsvHsCode/CSV/RowParser.php
```diff
@@ -82,8 +85,25 @@ private function getHsCode(array $rowData, ColumnResolver $columnResolver)
     {
         $hsCode = $columnResolver->getColumnValue(ColumnResolver::COLUMN_HSCODE, $rowData);
         if ($hsCode === '') {
-            $hsCode = '*';
+            $hsCode = null;
```
lizzyWithReach, line 104: why was  this change made? It's not super apparent from the other changes here so I wanted to check

Reach/Payment/Model/ResourceModel/Carrier/CsvHsCode/DataHashGenerator.php
```diff
@@ -17,6 +17,7 @@ public function getHash(array $data)
         $id = $data['id'];
         $sku = $data['sku'];
         $hsCode = $data['hs_code'];
-        return sprintf("%s-%d-%s-%F", $id, $sku, $hsCode);
+        $countryOfOrigin = $data['country_of_origin'];
+        return sprintf("%s-%d-%s-%F", $id, $sku, $hsCode, $countryOfOrigin);
```
lizzyWithReach: why is sprintf have %F for the countryOfOrigin value? I would have expected %s or at least some difference to the format since there's an extra field being printed

Reach/Payment/Model/DutyCalculator.php
```diff
@@ -342,6 +352,21 @@ protected function getHsCode($sku)
         return null;
     }
 
+        /**
```
lizzyWithReach: while you're  in these files, can you fix the spacing on this line? Small thing haha

lizzyWithReach (changes requested):
Just had a few comments, let me know when you get a chance to review

Reach/Payment/Model/ResourceModel/Carrier/CsvHsCode/DataHashGenerator.php
cstech: sprintf("%s-%s-%s-%", $id, $sku, $hsCode, $countryOfOrigin);
This returns everything with hypens as delimeters but yet if a sku contains a hypen then it may cause a problem. 
I still don't know where or what is using this function.

Reach/Payment/Model/ResourceModel/Carrier/CsvHsCode/CSV/RowParser.php
cstech: This change was made because when there was no hs code then it inserted an asterisk into the database.  This caused my check for a value to be true.  so if it was populated with an asterisk it wouldn't go to the default hs code of the merchant that I added in earlier pull requests

cstech:
Fixed issues will create new pull request

PR #21, opened 2019-11-12
Mag 11

cstech:
Did the recommended suggestions for importing of country of origin through plugin.
The default recommendation flags are included in this pull request.


cstech:
Found a few whitespace errors. Will fix and then do a new pull request.

PR #23, opened 2019-11-19
Moved files to work for packagist composer package 

cstech:
Moved files to work for packagist composer package file placement expectations


lizzyWithReach:
Since the original files are retained in the Reach/Payment folder and this is just for testing, I'm approving this PR. But when we do the final move we will need to take a look at the namespace declaration since it won't be 100% accurate (it includes Reach/Payment in the namespace path) and double check any other paths that exist in the code.

PR #24, opened 2019-11-25
Mag 23

cstech:
Made country submit to the API without a region

### DIFF
--- a/Model/DutyCalculator.php
+++ b/Model/DutyCalculator.php
@@ -132,7 +132,7 @@ class DutyCalculator implements \Reach\Payment\Api\DutyCalculatorInterface
                 return $this->response;
             }
            
-            if (empty($address->getCountryId()) || empty($address->getRegionCode())) {
+            if (empty($address->getCountryId())) {
                 $this->response->setSuccess(true);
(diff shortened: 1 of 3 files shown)


Model/DutyCalculator.php
```diff
@@ -132,7 +132,7 @@ public function getDutyandTax($cartId, $shippingCharge, $shippingMethodCode, $sh
                 return $this->response;
             }
            
-            if (empty($address->getCountryId()) || empty($address->getRegionCode())) {
```
lizzyWithReach, line 135: I don't understand how this works, if the country is empty then we got the duties and taxes? And this is an OR statement, so it should have worked before right?

lizzyWithReach (changes requested):
I'm just not sure why this code change works, if you could provide more context that would be great. I'm guessing I'm missing something

cstech:
Going to remove block of code

Model/DutyCalculator.php
lizzyWithReach: Clarified with Ervin, if the country is empty (which it won't ever be) then the duties would be reset to 0 and we would never call DHL. That's what this block is doing

PR #27, opened 2019-12-02
Mag 14

BronzePointKrel:
Hello Liz. The majority of changes are with Data/Reach.php

Let me know if you have any questions.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -14,11 +14,13 @@ class Data extends AbstractHelper
 
     const API_URL = 'https://checkout.gointerpay.net/v2.19/';
     const SANDBOX_API_URL = 'https://checkout-sandbox.gointerpay.net/v2.19/';
-    
+
     const DHL_API_URL = 'https://api.dhlecommerce.com/';
     const DHL_SANDBOX_API_URL = 'https://api-sandbox.dhlecommerce.com/';
+
     const DHL_ENABLE = 'reach/dhl/enable';
     const DHL_DUTY_LABEL = 'reach/dhl/duty_label';
+
     const DHL_DUTY_ALLOW_SPECIFIC = 'reach/dhl/allowspecific';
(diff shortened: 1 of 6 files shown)


Helper/Data.php
```diff
@@ -156,67 +199,77 @@ public function canAllowMultiPleCurrency()
     }
 
     /**
-     * Get currencly allowed for specific countries config
+     * Get currently allowed for specific countries config  REACH_SPECIFIC_PATH - Function 07
```
lizzyWithReach: should be currency, not currently

Helper/Data.php
```diff
@@ -156,67 +199,77 @@ public function canAllowMultiPleCurrency()
...
      *
      * @return boolean
      */
     public function allowCurrencySpeicifcCountry()
```
lizzyWithReach: can you correct this typo too? allowCurrencySpecificCountry

BronzePointKrel: No problem. There are spelling errors all over the place. Thankfully, most of them are in comments but sometimes (like this one) in function and method names.

Helper/Data.php
```diff
@@ -86,15 +93,51 @@ public function getReachConfig($code, $storeId = null)
         return $this->getConfigValue(self::XML_PATH_REACH .'reach_gointerpay/'. $code, $storeId);
     }
 
-     /**
-     * Check Reach Enabled
+    /**
+     * Get Currency Options value REACH_CURRENCY_SWITCH - Function 06
```
lizzyWithReach: out of curiosity, what's up with the Function numbers?

BronzePointKrel: I'll do another update and remove that. I numbered them and then checked them off as I went through and tested each function. Wanted to make sure that I did a complete cart and purchase while watching those checkboxes.

lizzyWithReach (commented):
Just had a couple of notes, if you could review the comments I've left

lizzyWithReach:
You got all of the comments handled haha thank you! I'll merge this in

PR #28, opened 2019-12-03
Use correct nomenclature in backend names

BronzePointKrel:
This update corrects the nomenclature used in the source code. Version number has been bumped up in anticipation of possible new release.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -36,9 +36,9 @@ class Data extends AbstractHelper
     const DHL_SPECIFIC_COUNTRY_PATH = "payment/reach_payment/reach_dhl/specificcountry";
     const DHL_OPT_SPECIFIC          = "payment/reach_payment/reach_dhl/optional_allowspecific";
     const DHL_OPT_SPECIFIC_COUNTRY  = "payment/reach_payment/reach_dhl/optional_specificcountry";
-    const DHL_KEY_PATH              = "payment/reach_payment/reach_dhl/dhl_key";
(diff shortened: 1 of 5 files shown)


Helper/Data.php
```diff
@@ -36,9 +36,9 @@ class Data extends AbstractHelper
     const DHL_SPECIFIC_COUNTRY_PATH = "payment/reach_payment/reach_dhl/specificcountry";
     const DHL_OPT_SPECIFIC          = "payment/reach_payment/reach_dhl/optional_allowspecific";
     const DHL_OPT_SPECIFIC_COUNTRY  = "payment/reach_payment/reach_dhl/optional_specificcountry";
-    const DHL_KEY_PATH              = "payment/reach_payment/reach_dhl/dhl_key";
+    const DHL_CLIENT_ID_PATH        = "payment/reach_payment/reach_dhl/dhl_client_id";
     const DHL_SHIPPING_PATH         = "payment/reach_payment/reach_dhl/applicable_shipping";
-    const DHL_SECRET_PATH           = "payment/reach_payment/reach_dhl/dhl_api_secret";
+    const DHL_CLIENT_SECRET           = "payment/reach_payment/reach_dhl/dhl_client_secret";
```
lizzyWithReach: should this be DHL_CLIENT_SECRET_PATH ? just to stay consistent.  Although if that's the case, we should probably fix DHL_OPT_SPECIFIC and DHL_OPT_SPECIFIC_COUNTRY, what do you think?

lizzyWithReach (commented):
Just had a little question on keeping the PATH consts consistent

lizzyWithReach (approved):
Looks good!

PR #29, opened 2019-12-09
Correct the configuration variable name.

BronzePointKrel:
There was a single line bug in Helper/Data.php which this corrects.

The minor version number was bumped up as well. New version number is 1.2.0

--ryan


lizzyWithReach (approved):
Sorry for not catching that! Thank you for fixing

PR #30, opened 2019-12-10
Decrypt api key before using

BronzePointKrel:
For security, Magento stores the Reach API key encrypted in the database.

Due to a bug, the module was attempting to process payment using the encrypted API key. This code fixes that bug so that the key is unencrypted temporarily before attempting to process payment with Reach.


lizzyWithReach (approved):
All good

PR #32, opened 2019-12-19
Use proper API endpoint based on sandbox checkbox in admin config

BronzePointKrel:
Was able to figure out a non-cookie solution using an already existing JS config data object.

### DIFF
--- a/Model/ReachConfigProvider.php
+++ b/Model/ReachConfigProvider.php
@@ -62,7 +62,7 @@ class ReachConfigProvider implements ConfigProviderInterface
         $config['reach']['dhl_quote']=$this->getDhlQuote();
         $config['reach']['fingerprint_url']=$this->getFingerPrintUrl();
         $config['reach']['dhl_label']=$this->reachHelper->getDhlDutyLabel();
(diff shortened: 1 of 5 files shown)


Model/ReachConfigProvider.php
```diff
@@ -62,6 +62,7 @@ public function getConfig()
         $config['reach']['dhl_quote']=$this->getDhlQuote();
         $config['reach']['fingerprint_url']=$this->getFingerPrintUrl();
         $config['reach']['dhl_label']=$this->reachHelper->getDhlDutyLabel();
+        $config['reach']['api_url']=$this->reachHelper->getStashApiUrl();
```
lizzyWithReach: can this be stash_url instead of api_url? To be less confusing

lizzyWithReach (changes requested):
Just a small note, otherwise this makes sense and I'm happy we could avoid using a cookie.

BronzePointKrel:
Updated with better aligned variable name.

lizzyWithReach (approved):

PR #33, opened 2019-12-20
Mag 17

BronzePointKrel:
Hello Lizzy. I am waiting for production credentials to finish testing this code out. Please hold on code review until I am able to verify locally with those credentials. Thank you.

### DIFF
--- a/etc/adminhtml/system.xml
+++ b/etc/adminhtml/system.xml
@@ -81,7 +81,7 @@
                         <source_model>Magento\Directory\Model\Config\Source\Country</source_model>
                             <can_be_empty>1</can_be_empty>
                     </field>
-                    <field id="dhl_key" showInDefault="1" showInStore="1" showInWebsite="1" sortOrder="70" translate="label" type="text">
+                    <field id="dhl_client_id" showInDefault="1" showInStore="1" showInWebsite="1" sortOrder="70" translate="label" type="text">
                         <label>DHL Client ID</label>
                         <depends>
                             <field id="active">1</field>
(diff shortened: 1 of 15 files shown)


etc/adminhtml/system.xml
```diff
@@ -110,9 +110,30 @@
                             <field id="active">1</field>
                         </depends>
                     </field>
-                    <field id="pricing_strategy" showInDefault="1" showInStore="1" showInWebsite="1" sortOrder="100" translate="label" type="select">
```
lizzyWithReach, line 137: does this mean that this setting will only be shown in the default config and not the store/website config? I'm not sure we want to restrict it to only the default view but I wanted to confirm first before checking with DHL

etc/adminhtml/system.xml
```diff
@@ -128,16 +149,14 @@
                         <depends>
                             <field id="active">1</field>
                         </depends>
-                        <config_path>reach/dhl/export_csv_hs_code</config_path>
                     </field>
-                    <field id="import_csv_hs_code" translate="label" type="Reach\Payment\Block\Adminhtml\Form\Field\Import" showInStore="1" sortOrder="130" showInDefault="1" showInWebsite="1" >
```
lizzyWithReach, line 159: This one too, will this section: ShowInStore="1" being removed mean that it won't be shown in the store config?

etc/adminhtml/system.xml
```diff
@@ -110,9 +110,30 @@
...
                         <label>DHL Pricing Strategy</label>
-                        <source_model>Reach\Payment\Model\Config\Source\Pricing</source_model>
+                        <source_model>Reach\Payment\Model\Config\Source\PricingStrategy</source_model>
+                        <depends>
+                            <field id="active">1</field>
+                        </depends>
+                    </field>
+                    <field id="clearance_mode" showInDefault="1" sortOrder="100" translate="label" type="select">
+                        <label>Clearance Mode</label>
+                        <source_model>Reach\Payment\Model\Config\Source\ClearanceMode</source_model>
```
lizzyWithReach, line 122: just to check, is the default value determined by the sort order? Or is there a way to explicitly set what the default value should be? (one of the requirements is that "Courier" is the default value for ClearanceMode for example)

lizzyWithReach (changes requested):

PR #34, opened 2020-01-08
Mag 45

BronzePointKrel:
- Revert back to 1.0.5 revision.
- Includes DHL identified issue about Line 71 in Model/Reach.php
- Includes Docker on MacOS edit.
- Bumps version number to 2.0.0

Tested on my local machine, including:
- Credit Card with sample data.
- Paypal with redirection to Paypal purchase
- Cash/Check

### DIFF
--- a/Model/Currency.php
+++ b/Model/Currency.php
@@ -208,6 +208,6 @@ class Currency extends \Magento\Framework\Model\AbstractModel
      */
     protected function checkLocalIP($ip)
     {
-        return in_array($ip, ['localhost','127.0.0.1']);
+        return in_array($ip, ['localhost','127.0.0.1','172.19.0.1']);
     }
 }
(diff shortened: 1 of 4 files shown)


Model/Currency.php
```diff
@@ -208,6 +208,6 @@ protected function getConsumerIp()
      */
     protected function checkLocalIP($ip)
     {
-        return in_array($ip, ['localhost','127.0.0.1']);
+        return in_array($ip, ['localhost','127.0.0.1','172.19.0.1','172.25.0.7','172.25.0.1']);
```
lizzyWithReach, line 215: could you just add a comment indicating what the extra IP addresses are for? I think localhost and 127.0.0.1 are self-explanatory, but it would help to have right here what the other IPs are for.

lizzyWithReach (approved):

PR #35, opened 2020-01-10
Mag 46 - Re-implement /stash hardcoded url fix

BronzePointKrel:
As a Magento merchant I want to use the Reach plugin in production so that I can use the plugin with actual customers.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -13,7 +13,7 @@ class Data extends AbstractHelper
 {
 
     const API_URL = 'https://checkout.gointerpay.net/v2.19/';
-    const STASH_URL = 'https://stash.gointerpay.net/v2.19/';
+    const STASH_URL = 'https://stash.gointerpay.net/';
     const STASH_SANDBOX_URL = 'https://stash-sandbox.gointerpay.net/';
     const SANDBOX_API_URL = 'https://checkout-sandbox.gointerpay.net/v2.19/';
 
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -13,8 +13,10 @@ class Data extends AbstractHelper
 {
 
     const API_URL = 'https://checkout.gointerpay.net/v2.19/';
+    const STASH_URL = 'https://stash.gointerpay.net/v2.19/';
+    const STASH_SANDBOX_URL = 'https://stash-sandbox.gointerpay.net/';
     const SANDBOX_API_URL = 'https://checkout-sandbox.gointerpay.net/v2.19/';
-    
+
     const DHL_API_URL = 'https://api.dhlecommerce.com/';
     const DHL_SANDBOX_API_URL = 'https://api-sandbox.dhlecommerce.com/';
     const DHL_ENABLE = 'reach/dhl/enable';
@@ -58,7 +60,21 @@ class Data extends AbstractHelper
         $this->_enc = $enc;
         parent::__construct($context);
     }
-    
+
+    /**
+     * Get Reach Stash API URL
+     *
+     * @return string
+     */
+    public function getStashApiUrl()
+    {
+        if ($this->getConfigValue(self::CONFIG_API_MODE)) {
+            return self::STASH_SANDBOX_URL;
+        } else {
+            return self::STASH_URL;
+        }
+    }
+
     /**
      * getConfigValue
      * gets config Value for frontend
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -43,6 +43,8 @@ class Data extends AbstractHelper
 
     const CONFIG_CC_OPEN_ONCTRACT = 'payment/reach_cc/allow_open_contract';
 
+    const SANDBOX_MODE = 1;
+
     /**
      * Constant for payment
      */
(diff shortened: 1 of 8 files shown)


Model/ReachConfigProvider.php
```diff
@@ -61,6 +61,7 @@ public function getConfig()
         $config['reach']['dhl_quote']=$this->getDhlQuote();
         $config['reach']['fingerprint_url']=$this->getFingerPrintUrl();  
         $config['reach']['dhl_label']=$this->reachHelper->getDhlDutyLabel();
+        $config['reach']['api_url']=$this->reachHelper->getStashApiUrl();
```
lizzyWithReach: I would prefer for this to be ['reach']['stash_url'] or 'stash_api_url' rather than just 'api_url' since that could get confusing, just a naming thing

view/frontend/web/js/view/payment/stash.js
```diff
@@ -120,7 +120,7 @@ define([
          */
         _getStash: function (stash_id) {                    
             var postData ={};
-            var url = "https://stash-sandbox.gointerpay.net/"+stash_id;
+            var url = window.checkoutConfig.reach.api_url + stash_id;
```
lizzyWithReach: see previous comment around changing api_url to stash_url or stash_api_url. I know in context it can be assumed that this is a stash url but it would still be preferable to be as explicit as possible.

lizzyWithReach (changes requested):
Hey Ryan, everything looked good other than a small note on the labeling of the stash api url. Let me know if this change is possible

Model/ReachConfigProvider.php
suraiyavic: We had a team discussion on Slack about this and I second Lizzy.

Model/DutyCalculator.php
```diff
@@ -132,10 +132,28 @@ public function getDutyandTax($cartId, $shippingCharge, $shippingMethodCode, $sh
                 return $this->response;
             }
             $accessToken = $this->getDhlAccessToken();
-           
+
+            // Enable lines with logger function to turn on logging for debugging.
+            // $this->_logger->debug('---------------- getDutyandTax - START OF REQUEST----------------');
```
suraiyavic, line 137: It would be a better practice for us not to comment out logging statement but to change log level based on which logging would be on or off.  The comment is relevant to multiple places in the commit. Thanks.

Helper/Data.php
```diff
@@ -58,7 +60,21 @@ public function __construct(
...
+
+    /**
+     * Get Reach Stash API URL
+     *
+     * @return string
+     */
+    public function getStashApiUrl()
+    {
+        if ($this->getConfigValue(self::CONFIG_API_MODE)) {
+            return self::STASH_SANDBOX_URL;
```
suraiyavic, line 74: From the conditional block and what is inside for sure we sense that when the returned value is true we are considering SANDBOX. Checking it against something like SANDBOX_MODE (holding a value equivalent to  true or false) or something equivalent would be cleaner/explicit.

view/frontend/web/js/view/payment/stash.js
suraiyavic: I second Lizzy. Thanks.

suraiyavic (commented):
As you see Ryan I have two additional recommendations/comments on top of what Lizzy added. Thanks.

Helper/Data.php
BronzePointKrel: I'm afraid I don't really understand your comment. I followed the existing pattern that the rest of the file was written around.

Model/DutyCalculator.php
BronzePointKrel: I agree with your statement but my lines are much more verbose than would ever be needed. I would prefer to keep them commented out for now so they can be easily found and remove them completely when we have a better working version of the project. I literally dump every variable out and create a ton of noise in the logs that were necessary to figure out how things work.

Model/ReachConfigProvider.php
BronzePointKrel: Variable name has been changed.

view/frontend/web/js/view/payment/stash.js
BronzePointKrel: Variable name has been changed.

Helper/Data.php
suraiyavic: if ($this->getConfigValue(self::CONFIG_API_MODE) ==  SANDBOX_MODE) something like that

BronzePointKrel:
I believe the requested changes have been addressed. @lizzynibali are you able to merge this into Master? Thank you.

lizzyWithReach (approved):

Model/DutyCalculator.php
suraiyavic: If the team prefers to keep it this way for now then I am okay with that Ryan.

That said in general this type of things keep accumulating over time which makes me hesitant to ignore this. 

Thanks.

suraiyavic (approved):
As mentioned in slack: I am very hesitant with the way logging is handled at present ... better practice is to turn logging on or off by setting log level higher or lower depending on whether it is dev/qa or prod ... not by commenting log statements out. But as per the discussion we are leaving it the way it is. Thanks!

PR #36, opened 2020-01-14
Mag 42 - Double check DHL Enabled at the Default and Store levels

BronzePointKrel:
Previously MAG-34

### DIFF
--- a/view/frontend/web/template/shipping.html
+++ b/view/frontend/web/template/shipping.html
@@ -59,7 +59,7 @@
 
                 <render args="shippingMethodListTemplate"/>
                 
-                <div class="reach-tax-duty" data-bind="fadeVisible: dutyCharges() > 0 && getReachEnabled() != 0">
+                <div class="reach-tax-duty" data-bind="fadeVisible: dutyCharges() > 0 && getReachEnabled() != 0 && getDHLEnabled() != 0">
                     <input 
                     type="checkbox"
                     class="checkbox-inline" ,                    
--- a/view/frontend/web/template/shipping.html
+++ b/view/frontend/web/template/shipping.html
@@ -59,7 +59,7 @@
 
                 <render args="shippingMethodListTemplate"/>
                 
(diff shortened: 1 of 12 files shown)


view/frontend/web/template/shipping.html
```diff
@@ -59,7 +59,7 @@
 
                 <render args="shippingMethodListTemplate"/>
                 
-                <div class="reach-tax-duty" data-bind="fadeVisible: dutyCharges() > 0 && getReachEnabled() != 0">
+                <div class="reach-tax-duty" data-bind="fadeVisible: dutyCharges() > 0 && getReachEnabled() != 0 && getDHLEnabled() != 0">
```
lizzyWithReach: the only issue here is that the duties and taxes checkbox should show up if DHL is enabled, but it doesn't matter if Reach is enabled. They should work independently.

lizzyWithReach: (by "they" I mean the Reach vs DHL functionality)

lizzyWithReach (changes requested):

view/frontend/web/template/shipping.html
suraiyavic: I do not know enough about the business rules to check/comment about the correctness here. So I would go with what Lizzy said.

I have to read/revisit the older meeting notes available in our system. But would appreciate if someone points me to any better source of information that they can think of/know of. Thanks.

suraiyavic (commented):
Actually as I think about it Lizzy's comment about one independent (duty label can be enabled even if reach is not) of the other makes sense.

view/frontend/web/template/shipping.html
BronzePointKrel: Updated code to reflect requested change.

BronzePointKrel:
I believe all issues are now resolved.

lizzyWithReach (approved):

PR #37, opened 2020-01-14
MAG-16 - Make DHL Recommended fields configurable

BronzePointKrel:
Was previously MAG-17

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -13,7 +13,7 @@ class Data extends AbstractHelper
 {
 
     const API_URL = 'https://checkout.gointerpay.net/v2.19/';
-    const STASH_URL = 'https://stash.gointerpay.net/v2.19/';
+    const STASH_URL = 'https://stash.gointerpay.net/';
     const STASH_SANDBOX_URL = 'https://stash-sandbox.gointerpay.net/';
     const SANDBOX_API_URL = 'https://checkout-sandbox.gointerpay.net/v2.19/';
 
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -42,10 +42,10 @@ class Data extends AbstractHelper
 
     const CONFIG_CC_OPEN_ONCTRACT = 'payment/reach_cc/allow_open_contract';
 
-    CONST DHL_PRICING_STRATEGY_PATH = "payment/reach_payment/reach_dhl/pricing_strategy";
-    CONST DHL_CLEARANCE_MODE_PATH   = "payment/reach_payment/reach_dhl/clearance_mode";
-    CONST DHL_END_USE_PATH          = "payment/reach_payment/reach_dhl/end_use";
-    CONST DHL_TRANSPORT_MODE_PATH   = "payment/reach_payment/reach_dhl/transport_mode";
+    const DHL_PRICING_STRATEGY_PATH = "payment/reach_payment/reach_dhl/pricing_strategy";
+    const DHL_CLEARANCE_MODE_PATH   = "payment/reach_payment/reach_dhl/clearance_mode";
+    const DHL_END_USE_PATH          = "payment/reach_payment/reach_dhl/end_use";
+    const DHL_TRANSPORT_MODE_PATH   = "payment/reach_payment/reach_dhl/transport_mode";
     const SANDBOX_MODE = 1;
 
     /**
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -30,7 +30,6 @@ class Data extends AbstractHelper
     const DHL_API_SECRET = 'reach/dhl/api_secret';
     const DHL_ITEM_SELLER = 'reach/dhl/item_seller';
     const DHL_PICKUP_ACCOUNT = 'reach/dhl/pickup_account';
-    const DHL_PRICIING_STRATEGY = 'reach/dhl/pricing_strategy';
     const DHL_DEFAULT_HS_CODE = 'reach/dhl/default_hs_code';
 
     const CONFIG_REACH_ENABLED = 'reach/global/active';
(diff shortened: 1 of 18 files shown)


Helper/Data.php
```diff
@@ -43,6 +42,10 @@ class Data extends AbstractHelper
 
     const CONFIG_CC_OPEN_ONCTRACT = 'payment/reach_cc/allow_open_contract';
 
+    CONST DHL_PRICING_STRATEGY_PATH = "payment/reach_payment/reach_dhl/pricing_strategy";
```
lizzyWithReach: just curious, is there a reason these CONST's are all caps? The other ones aren't

BronzePointKrel: Fixed 🙃

suraiyavic (commented):
Please make sure that overall correct default values are set (as per the acceptance criteria of JIRA MAG-16). Thanks.

etc/adminhtml/system.xml
```diff
@@ -135,11 +134,31 @@
...
+                    <field id="clearance_mode" showInDefault="1" showInStore="1" showInWebsite="1" sortOrder="100" translate="label" type="select">
+                        <label>Clearance Mode</label>
+                        <source_model>Reach\Payment\Model\Config\Source\ClearanceMode</source_model>
+                        <depends>
+                            <field id="active">1</field>
+                        </depends>
+                    </field>
+                    <field id="end_use" showInDefault="1" showInStore="1" showInWebsite="1" sortOrder="100" translate="label" type="select">
+                        <label>End Use</label>
+                        <source_model>Reach\Payment\Model\Config\Source\EndUse</source_model>
```
suraiyavic, line 158: Please kindly make sure appropriate default values  (as per JIRA MAG-16) are set for all these fields. Thanks!

Helper/Data.php
```diff
@@ -104,6 +107,43 @@ public function getReachConfig($code, $storeId = null)
...
+     * @return void
+     */
+    public function getTransportMode() {
+        return $this->getConfigValue(self::DHL_TRANSPORT_MODE_PATH);
+    }
+
+    /**
+     * Get End Use DHL_END_USE_PATH
+     *
+     * @return string|null
```
lizzyWithReach: since there will always be a default value, is it true that these functions may return null?

lizzyWithReach (changes requested):
The only note remaining is to make sure that a default is set for each of these configurable options. What the default should be is written out in MAG-16, but let us know if setting a default isn't possible (or something like that is happening)

Helper/Data.php
BronzePointKrel: Correct. These values are from a drop down list that must contain a value from the list. Null would be applicable if text field or text entry box.

etc/adminhtml/system.xml
BronzePointKrel: Default values set. Can be confirmed in default.xml

BronzePointKrel (commented):
I believe all requested changes have been applied. Tested locally to confirm no breaking changes.

Helper/Data.php
```diff
@@ -43,6 +42,11 @@ class Data extends AbstractHelper
 
     const CONFIG_CC_OPEN_ONCTRACT = 'payment/reach_cc/allow_open_contract';
 
+    const DHL_PREF_TARIFFS          = "payment/reach_payment/reach_dhl/pref_tariffs";
```
lizzyWithReach, line 45: do we need a get function for this config value?

BronzePointKrel: > Let me be bit more specific: with various values available in the toOptionArray as well as corresponding 1 would correct default value be set at the beginning after page rendering? This comment is about each of qualifiesForPreferentialTariffs, pricingStrategy, clearanceMode, endUse, transportMode. Thank you!

Not 100% following. During system:upgrade, the Magento system will use 'defaults.xml' to create the database entries if they do not yet currently exist and fill the values with the default values.

When the page loads (what you seem to be asking), it will display those default values until they are changed by the user.

etc/config.xml
```diff
@@ -1,34 +1,38 @@
...
-              <title>Paypal (Reach)</title>
-              <allowspecific>0</allowspecific>
-            </reach_paypal>  
+                <reach_dhl>
+                    <enable>0</enable>
+                    <duty_label>Pay Tax and Duties</duty_label>
+                    <pricing_strategy>MAXIMUM</pricing_strategy>
+                    <allowspecific>0</allowspecific>
+                    <pref_tariffs>1</pref_tariffs>
+                    <pricing_strategy>MAXIMUM</pricing_strategy>
```
lizzyWithReach: pricing strategy is here twice (see line 12)

Helper/Data.php
suraiyavic: Escaped the html tags to match Github markdown and this is how it looks like now:

Let me be bit more specific: with various values available in the toOptionArray as well as corresponding \<field id="active">1\</field> would correct default value be set at the beginning after page rendering? This comment is about each of qualifiesForPreferentialTariffs, pricingStrategy, clearanceMode, endUse, transportMode. Thank you! 

Does that help?

lizzyWithReach (changes requested):
a comment on get function for preferential tariffs and just a duplicate line in the config.xml

Helper/Data.php
BronzePointKrel: Again, not following what you are asking. The XML snippet you quoted is saying that this field should not be editable unless the parent is enabled. For example, if DHL is disabled then all sub-items will be grayed out in the UX.

suraiyavic: If it is behaving/setting things as intended (as per your test before commit) then I am good. Thanks!

lizzyWithReach (approved):

PR #40, opened 2020-01-16
Mag 64 - Use string instead of 0 or 1

BronzePointKrel:
Also bumps version number.


Model/DutyCalculator.php
```diff
@@ -320,7 +319,11 @@ protected function prepareRequest($freightCharge, $shippingAddress)
                 if (!$itemData['countryOfOrigin']) {
                     $itemData['countryOfOrigin'] = $request['senderAddress']['country'];
                 }
-                $itemData['qualifiesForPreferentialTariffs']=true;
+                if ($this->reachHelper->getPrefTariffs() == 1 ) {
```
lizzyWithReach, line 322: does Magento restrict the Yes/No field to 1/0? or could we map those values to True/False and avoid this if statement

lizzyWithReach (approved):
just had a little q, but I'm approving the PR because this looks right and I understand where it's coming from. But if you could reply to the q that would help

PR #41, opened 2020-01-20
Mag 32 - Send through the RateOfferId with all /checkout requests

BronzePointKrel:
The JIRA ticket (https://withreach.atlassian.net/browse/MAG-32) contains a comment that shows how to verify that this code is working.


suraiyavic (commented):
Seems to be a fairly simple set of changes: adding a new request param at two places and adjusting Cron schedule (in addition to some logging). I would love to hear Lizzy's view on guaranteed vs spot aspect before approving. Thanks.

lizzyWithReach (approved):
Looks good

PR #42, opened 2020-01-23
Fix ClearanceMode to have a no selection option

BronzePointKrel:
I have to put a label of some sort into the drop-down, so I went with 'No Selection'.

If both the value and label are blank, then the UX for the drop down only gives the other two values as options.

This label can be changed if the team wants a different label.


lizzyWithReach (approved):
It would be good to have DHL take a look at this once it's pushed through, to make sure it looks the way they expect it to. But according to the ticket this should do the trick

suraiyavic (approved):
Minor change. Looks okay. Thanks.

BronzePointKrel:
Looks like this is approved by both of you. I am going to merge this into master so I can move on to the next ticket. Thoughts?

lizzyWithReach:
Sounds good! Thanks Ryan

PR #43, opened 2020-01-24
MAG-66 - Incorrect DHL Defaults for Qualifies for Preferential Tariffs & DHL Pricing Strategy

BronzePointKrel:
Also fixes MAG-67


lizzyWithReach (approved):
looks good, I guess having the dhl tags in the config.xml didn't work for setting defaults?

suraiyavic (approved):
I am not sure about the change in the etc/config.xml and was hesitant to approve it seeing what happened in the recent past with similar changes.
Discussed with Ryan about it on Slack but the answer was not comprehensive/very logical to me. 
Anyway approving it as per my discussion with Lizzy.
Thanks.

BronzePointKrel:
I've closed this pull request for the time being. It can be re-opened after Suraiya finishes her investigation into the config xml files.

PR #44, opened 2020-01-27
[MAG-70] Reducing redundant calls to DHL Duty API

suraiyavic:
This work tries to reduce redundant/extra calls to DHL duty calculation service.

In pre-existing extension code as is the following activities of probable buyer/customer in Magento store/front use to cause multiple/redundant calls (DHL duty calculation).

- Changes in digits in postal code, 

- Changes in state/division (even when it is not relevant for calculating duties for the shipping address under consideration), 

-  Validation error in user name email etc. used.

-  When apply duty & tax checkbox is selected or deselected from the checkout page

- When the 'next' button is pressed on the checkout page. 

Tried to fix the problem leveraging Magento Checkout Session as well as 'state' of DHL Quote stored in the database. A separate document will be supplied highlighting how it was tested, test cases suggestions, limitations of the current solution etc.

Committed some extraneous blank lines. Would be more than happy to remove.

### DIFF
--- a/Model/DutyCalculator.php
+++ b/Model/DutyCalculator.php
@@ -116,84 +116,347 @@ class DutyCalculator implements \Reach\Payment\Api\DutyCalculatorInterface
         $this->_logger = $logger;
     }
 
+
+    /**
+     * Sets quote and checkout Session state based on whether duty and tax is applicable or not
+     * @param Magento\Quote\Api\Data\AddressInterface $address
+     * @param bool $apply  - whether duty is applicable or not
+     *
+     */
+    public function handleTaxApplicability($address, $apply)
+    {
+
+        $this->_logger->debug('In handleTaxApplicability method');
+        $this->_logger->debug($this->checkoutSession->getReachDuty());
+        #$apply = $this->checkoutSession->getApply();
+        $this->_logger->debug($address->getCountryId());
+        $this->_logger->debug($address->getRegionCode());
+        $this->_logger->debug($this->checkoutSession->getPrevCountry());
+        $this->_logger->debug($this->checkoutSession->getPrevRegion());
+        $this->_logger->debug($apply);
+        $quote = $this->checkoutSession->getQuote();
+        if ($apply || !$this->getIsOptional($address->getCountryId())) {
+           //$quote->setDuty($this->checkoutSession->getDuty());
+           $quote->setReachDuty($this->checkoutSession->getReachDuty());
+           $this->response->setIsOptional($this->getIsOptional($address->getCountryId()));
+           $this->response->setSuccess(true);
+           $this->response->setDuty($this->checkoutSession->getRechDuty());
+           $this->_logger->debug('In handleTaxApplicability method --- apply is true');
+        }
+        else {
+           $this->response->setIsOptional($this->getIsOptional($address->getCountryId()));
+           $this->response->setSuccess(true);
+           #as apply duty and tax(DT) is not selected duty and tax would not be used in total pricing
+           #so setting all relevant quote values to zero
+           //$quote->setDuty(0) ;
+           $quote->setReachDuty(0) ;
+           $this->response->setDuty($this->checkoutSession->getReachDuty()); //the DT checkbox and label  should still appear
+           //if this amount is more than 0 even though the user did not choose to apply it on
+           //computation of total landed cost/billing
+           $this->_logger->debug('In handleTaxApplicability method --- apply is false');
+
+        }
+        $quote->save();//this is needed so that different aspects related to a quote are available on other pages.
+        //More specifically saving quote in database is needed as apparently Magento session scopes are not application
+        // wide but code area specific (?)
+    }
+
+
+    /** handles when duty is not applicable (based on some configuration) for the country
+     * or shipping is not allowed for the country
+     * and saves proper state of quote so that the information is available beyond this page/area
+     * @param float $duty
+     * @param \Magento\Quote\Model\Quote $quote
+     */
+    public function handleNoDutyOrShippingCase($duty, $quote)
+    {
+        $this->response->setSuccess(true);
+        $this->response->setDuty($duty);
+        $this->checkoutSession->setReachDuty($duty);
+        $quote->setReachDuty($duty) ;
+        $this->_logger->debug('In duty or shipping not allowed section');
+        $quote->save();
+    }
+
+    /** Sets appropriate response (eventually be https response) values
+     * when state/province is (needed for proper duty calculation
+     * but) not yet specified
+     * @param float $duty
+     */
+    public function handleStateUnspecifiedCase($duty)
+    {
+       $quote = $this->checkoutSession->getQuote();
+       $this->response->setSuccess(true);
+       $this->response->setDuty($duty);
+       $this->_logger->debug('Special country where state selection is neccessary before initiating DHL API call;
+                but state is not selected anyway.');
+    }
+
+
+    /** Sets appropriate response (\Reach\Payment\Api\Data\DutyResponseInterface) values
+     *  when both state/province and country are specified
+     * @param Magento\Quote\Api\Data\AddressInterface $address
+     * @param bool $apply
+     */
+    public function handleCaseWithCountryAndStateSpecified($address, $apply)
+    {
+        $quote = $this->checkoutSession->getQuote();
+        $this->response->setDuty($quote->getReachDuty());
+        $this->handleTaxApplicability($address, $apply);
+        $this->_logger->debug('Country is specified but state is not selected (when both are needed); so we would'.
+            'prevent call to DHL API');
+    }
+
+
+    /** Uses already (recently) retrieved duty from DHL
+     * @param bool $apply
+     * @param \Magento\Quote\Model\Quote $quote
+     * @param Magento\Quote\Api\Data\AddressInterface $address
+     */
+    public function handleCaseWhereDutyIsAlreadyRetrieved($apply, $quote, $address)
+    {
+        if ($apply) {//user indicated that (s)he wants to apply duty and tax during calculation of billing
+
+            $quote->setReachDuty($this->checkoutSession->getReachDuty());
+            $this->response->setDuty($this->checkoutSession->getReachDuty());
+            $this->handleTaxApplicability($address, $apply);
+            $this->_logger->debug('User wants to apply the D&T. Value of apply ::' . $apply);
+            $quote->setBaseReachDuty($this->checkoutSession->getBaseReachDuty());
+            $quote->setReachDuty($this->checkoutSession->getReachDuty());
+            $quote->setDhlQuoteId($this->checkoutSession->getDhlQuoteId());
+            $quote->setDhlBreakdown($this->checkoutSession->getDhlBreakdown());
+            $this->_logger->debug($quote->getReachDuty());
+        }
+        else {////user indicated that (s)he does not  want to apply duty and tax during calculation of billing
+            //$quote->setDuty(0) ; #$this->checkoutSession->getDuty());
+            $quote->setReachDuty(0);
+            //should we reset DhlQuoteId and DhlBreakdown too so as not to break any reporting?
+            //I am assuming that we do
+            $quote->setBaseReachDuty(0);
+            $quote->setDhlQuoteId('');
+            $quote->setDhlBreakdown('');
+            $this->_logger->debug('User does not want to apply the D&T. Value of apply ::' . $apply);
+        }
+
+        $quote->save();
+
+        $this->response->setDuty($this->checkoutSession->getReachDuty());
+    }
+
+    /** Adjust returned duty value from DHL and additionally keep track of different data that came back from
+     * DHL API call and saves proper state of quote so that the information is available beyond this page/area
+     * @param float $duty
+     * @param Magento\Quote\Api\Data\AddressInterface $address
+     * @param bool $apply
+     * @param  Magento\Framework\App\Response $response
+     */
+    public function fillOutQuoteAndSessionUsingFeeReturned($duty, $address, $apply, $response)
+    {
+
+        //dealing with whether duty and tax related checkbox is selected or not
+        //or whether applying duty is a must for that country or not (a setting in magento admin
+        //panel)
+        //filling out both Quote and session state as appropriate
+        //DHL quoteID is already saved into session; so not doing it here again
+        $quote = $this->checkoutSession->getQuote();
+        $duty_adjusted = $this->priceCurrency->round($duty); //copied over pre-existing code
+        //but are we supposed to round always?
+        //should not that be based on corresponding admin setting?
+        $this->checkoutSession->setReachDuty($duty_adjusted);
+
+        if ($apply || !$this->getIsOptional($address->getCountryId())) {
+            //checkbox selection was duty should be applied
+            //or applying duty is a must for that country
+            $baseCurrency = $this->storeManager->getStore()->getBaseCurrency();
+            $rate = $baseCurrency->getRate($baseCurrency->getCode());
+            $baseDuty = $duty_adjusted / $rate;
+            $quote->setBaseReachDuty($baseDuty);
+            $this->checkoutSession->setBaseReachDuty($baseDuty);
+            $quote->setReachDuty($duty_adjusted);
+            $this->checkoutSession->setReachDuty($duty_adjusted);
+            $quote->setDhlQuoteId($response['quoteId']);
+            $quote->setDhlBreakdown(json_encode($response['feeTotals']));
+            $this->checkoutSession->setDhlBreakdown($quote->getDhlBreakdown());
+            $this->checkoutSession->setApply(true); #checkbox selection /corresponding passed value
+            # implies that duty should be applied
+            $this->_logger->debug('Apply block immediately after DHL call');
+
+        } else {
+            //checkbox selection or paramter passed indicates that duty should not be applied
+            $quote->setBaseReachDuty(0);
+            $quote->setReachDuty(0);
+            $this->checkoutSession->setBaseReachDuty(0);
+            $this->checkoutSession->setReachDuty($duty_adjusted);
+            $quote->setDhlQuoteId('');
+            $this->checkoutSession->setDhlQuoteId('');
+            $quote->setDhlBreakdown('');
+            $this->checkoutSession->setDhlBreakdown('');
+            $this->checkoutSession->setApply(false);
+            $this->_logger->debug('Do not Apply block immediately after DHL call');
+
+        }
+        $quote->save();
+        $this->response->setSuccess(true);
+        $this->response->setDuty($duty_adjusted);
+        $this->response->setIsOptional($this->getIsOptional($address->getCountryId()));
+    }
+
+    /**
+     * Deals with a case when DHL API call does not return a value for duty
+     * @param Magento\Framework\App\Response $response
+     */
+    public function fillOutQuoteAndSessionOnError($response)
+    {
+        $this->response->setSuccess(false);
+        $this->response->setDuty(0);
+        if (isset($response['message'])) { //error message from DHL duty and tax api call?
+            //if yes then do we want to call the API again?
+            //Assuming we do; we are resetting previous country and region value so that we can reenter the blcok to make
+            //DHL API call
+            $this->checkoutSession->setResponseErrorMessage($response['message']);
+            $this->response->setErrorMessage($response['message']);
+            $this->checkoutSession->setPrevCountry('');
+            $this->checkoutSession->setPrevRegion('');
+        }
+    }
+
+    /**
+     * Retrieves duty value for a country or (country, state) combo
+     * and set state of quote, response, session accordingly/as per the value returned
+     * @param $cartId
+     * @param Magento\Quote\Api\Data\AddressInterface $address
+     * @param float $shippingCharge
+     * @param float $duty
+     * @param bool $apply
+     * @param \Magento\Quote\Model\Quote $quote
+     */
+    public function callDHLDutyTaxApi($cartId, $address, $shippingCharge, $duty, $apply, $quote)
+    {
+        $accessToken = $this->getDhlAccessToken();
+
+
+        //Right moment to make a DHL api call for getting 'Duty and Tax' value
+        if ($accessToken && $accessToken != '') {
+            $this->checkoutSession->setCartId($cartId);
+            //storing selection of country and state to be able to compare it against new set of values
+            //when a potential buyer inputs more data
+            $this->checkoutSession->setPrevCountry($address->getCountryId());
+            $this->checkoutSession->setPrevRegion($address->getRegionCode());
+
+            $request = $this->prepareRequest($shippingCharge, $address);
+            $this->_logger->debug('---------------- Making DHL API call to get Duty and Tax ----------------');
+            $response = $this->getQuote($request, $accessToken);
+
+            $this->checkoutSession->setDhlQuoteId($response['quoteId']);
+
+            //fee value came from DHL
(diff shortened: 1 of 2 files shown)


Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+        $this->_logger->debug('In handleTaxApplicability method');
+        $this->_logger->debug($this->checkoutSession->getReachDuty());
+        $this->_logger->debug($address->getCountryId());
+        $this->_logger->debug($address->getRegionCode());
+        $this->_logger->debug($this->checkoutSession->getPrevCountry());
+        $this->_logger->debug($this->checkoutSession->getPrevRegion());
+        $this->_logger->debug($apply);
+        $quote = $this->checkoutSession->getQuote();
+
+        if ($apply || !$this->getIsOptional($address->getCountryId())) {
```
lizzyWithReach: these two logical cases follow a lot of the same logic with the same lines. Would it be clearer to set the ReachDuty to 0 if (!$apply || $this->getIsOptional($address->getCountryId)) and otherwise set the ReachDuty to $this->getIsOptional($address->getCountryId())  since that appears to be the only things that is different between the two cases?

But I might be reading this wrong, just wanted to check since it took me a bit to work out what was happening here.

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+           $quote->setReachDuty($this->checkoutSession->getReachDuty());
+           $this->response->setIsOptional($this->getIsOptional($address->getCountryId()));
+           $this->response->setSuccess(true);
+           $this->response->setDuty($this->checkoutSession->getRechDuty());
+           $this->_logger->debug('In handleTaxApplicability method --- apply is true');
```
lizzyWithReach: this debug comment is a bit misleading, right? Because either apply is true or no CountryId could be retrieved, the code could still run even if apply was false in this case

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+           $quote->setReachDuty(0) ;
+           $this->response->setDuty($this->checkoutSession->getReachDuty()); //the DT checkbox and label  should still appear
+           //if this amount is more than 0 even though the user did not choose to apply it on
+           //computation of total landed cost/billing
+           $this->_logger->debug('In handleTaxApplicability method --- apply is false');
+        }
+
+        $quote->save();//this is needed so that different aspects related to a quote are available on other pages.
+        //More specifically saving quote in database is needed as apparently Magento session scopes are not application
+        // wide but code area specific (?)
```
lizzyWithReach: @ryanjohnston could you confirm this so that our comment is accurate and factual going forward

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+    }
+
+
+    /** handles when duty is not applicable (based on some configuration) for the country
```
lizzyWithReach, line 163: would it be helpful to be more specific for what we mean by "some configuration"? Or would it be a bad ideal to tie this comment to configuration details that could change over time

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+    {
+        $this->response->setSuccess(true);
+        $this->response->setDuty($duty);
+        $this->checkoutSession->setReachDuty($duty);
+        $quote->setReachDuty($duty) ;
+        $this->_logger->debug('In duty or shipping not allowed section');
+        $quote->save();
+    }
+
+    /** Sets appropriate response (eventually be https response) values
```
lizzyWithReach: do we have a ticket to make this response https? If there's work we want to do my personal preference would be to make the ticket and potentially refer to it in the comment

Model/DutyCalculator.php
suraiyavic: Right. During debug message beautification (replacing old one before commit) this happened. :( Going to change and thank you.

Model/DutyCalculator.php
suraiyavic: To help with this a bit:
The information that we are manipulating in dutyCalculation.php is required in duty.php as well to show proper values of applicable duty. With the code as is that block is looking at quote info /state that is in the db not what I stored in session in dutyCalculation.php. To make sure proper state is available in duty.php this state save thing was done. Thanks.

suraiyavic: And the session was in dutyCalculation.php so that I do not have to make another call to retrieve duty from DHL. When user deselect the apply duty checkbox the duty is set to zero in db state but the value returned from DHL api call is saved in session to be able to reuse it if the user changes mind or when that page is getting entered repeatedly due to different events (mostly Magento ones). Instead of session this could be saved in database too. However that would need creation of new table and or fields in the database.

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+
+
+    /** Uses already (recently) retrieved duty from DHL
+     * @param bool $apply
+     * @param \Magento\Quote\Model\Quote $quote
+     * @param Magento\Quote\Api\Data\AddressInterface $address
+     */
+    public function handleCaseWhereDutyIsAlreadyRetrieved($apply, $quote, $address)
+    {
+        if ($apply) {//user indicated that (s)he wants to apply duty and tax during calculation of billing
```
lizzyWithReach, line 222: ohhh I didn't know what this variable meant until this comment, maybe we can have a better description of what the variable represents in every line that it shows up in the function comments. For example, line 210, 197, 123. Or a more descriptive variable name.

Brainstorming: userPayFLC, payDutiesAndTaxesChecked, applyFLC...

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+            $quote->setReachDuty($this->checkoutSession->getReachDuty());
+            $this->response->setDuty($this->checkoutSession->getReachDuty());
+            $this->handleTaxApplicability($address, $apply);
+            $this->_logger->debug('User wants to apply the D&T. Value of apply ::' . $apply);
```
lizzyWithReach, line 226: I love this debug statement, could we use the same thing (or something like it) on lines 142, 154?

Model/DutyCalculator.php
suraiyavic: https://mage2-blog.com/magento-2-get-session/ has some info on sessions. I should have a better link but not handy right now.

Adding checkout session in duty.php and then retrieving value from there could work too as another possible solution. But then do not know (without checking) where else that fix would break due to the state not being saved in db (so preventing propagation as needed). Significant time was gone chasing events from all directions.

Model/DutyCalculator.php
suraiyavic: It is a pre-existing check that I reused. This is the area where I am confused that I mentioned during the stand up today. I do not know what is a feature and what is a bug in this system - the ground truth is lacking which makes testing harder/(makes us chase many directions). The changes that I was doing in admin panel to test this area (in store) are either wrong or there is a bug in pre-existing code (dealing with admin setting). The decision during the meeting was to request you @lizzynibali for a demo/discussion of the system so that I can be very precise/correct with my description/ what I am doing. (Rephrasing the team recommendation in my own language ) Thanks.

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+    public function fillOutQuoteAndSessionUsingFeeReturned($duty, $address, $apply, $response)
+    {
+        //dealing with whether duty and tax related checkbox is selected or not
+        //or whether applying duty is a must for that country or not (a setting in magento admin
+        //panel)
+        //filling out both Quote and session state as appropriate
+        //DHL quoteID is already saved into session; so not doing it here again
+        $quote = $this->checkoutSession->getQuote();
+        $duty_adjusted = $this->priceCurrency->round($duty); //copied over pre-existing code
+        //but are we supposed to round always?
```
lizzyWithReach, line 263: you're right, it looks like this function by default rounds to zero decimal places which doesn't seem right https://www.geeksforgeeks.org/php-round-function/ I would have expected it to round to two decimal places so that it was a valid amount to charge but not to zero, that could alter the amount paid by quite a bit. This might be something we need a specific ticket for so we can see what DHL would prefer

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+        //should not that be based on corresponding admin setting?
+        $this->checkoutSession->setReachDuty($duty_adjusted);
+
+        if ($apply || !$this->getIsOptional($address->getCountryId())) {
+            //checkbox selection was duty should be applied
+            //or applying duty is a must for that country
```
lizzyWithReach, line 269: I didn't think any country had applying duty as a must, could you help me if I missed this requirement in the DHL API spec or somewhere else?

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+    /**
+     * Deals with a case when DHL API call does not return a value for duty
+     * @param Magento\Framework\App\Response $response
+     */
+    public function fillOutQuoteAndSessionOnError($response)
+    {
+        $this->response->setSuccess(false);
+        $this->response->setDuty(0);
+
+        if (isset($response['message'])) { //error message from DHL duty and tax api call?
```
lizzyWithReach, line 313: based on where this function is called I would also assume that this contains an error from the DHL duty and tax api call. I wonder if we should do different things based on the error received? Maybe this should be its own ticket to work out how to handle an error from the DHL API (for example, I wouldn't want to leave the checkbox available if we're getting big errors from DHL (400, something wrong with our API call at a fundamental level) but a 503 or something I'd be ok retrying.

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+            //if yes then do we want to call the API again?
+            //Assuming we do; we are resetting previous country and region value so that we can reenter the blcok to make
+            //DHL API call
+            $this->checkoutSession->setResponseErrorMessage($response['message']);
+            $this->response->setErrorMessage($response['message']);
+            $this->checkoutSession->setPrevCountry('');
+            $this->checkoutSession->setPrevRegion('');
```
lizzyWithReach, line 320: I see that only the session is being reset, should the quote be reset and saved as well?

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+
+            }
+        }
+    }
      /**
       * @inheritDoc
       */
     public function getDutyandTax($cartId, $shippingCharge, $shippingMethodCode, $shippingCarrierCode, $address, $apply = false)
     {
+        //Some comments/concerns which may be relevant to the reviewer
```
lizzyWithReach: these comments are really helpful, but would it be possible to add them in the ticket or as a part of the pull request rather than in the code as comments? Unless they were intended to be temporary until the review was complete and then edited out.

Model/DutyCalculator.php
```diff
@@ -116,84 +116,333 @@ public function __construct(
...
+                if (in_array($address->getCountryId(), $special_countries, true)) {
+                    if (($this->checkoutSession->getPrevRegion() == $address->getRegionCode())
+                        || !$address->getRegionCode()  //this check is redundant here
+                    ) {
+                        $this->handleCaseWithCountryAndStateSpecified($address, $apply);
+                        return $this->response;
                     }
+                }
+                else {//country selection did not change and we are assuming (as we do not have extra info handy)
+                    //for these countries D&T does not changes with change in state (double check with the business)
```
lizzyWithReach: Miguel can double check with DHL but the duties and taxes do change based on region, so we'll need to send another request in the case that the country is the same but the region is different.  We can't take the previously calculated duty amount if the region has changed

lizzyWithReach (changes requested):
There are a couple of comments that are more important than others haha let me know if I can highlight those over the others since I know I wrote a lot out.

I totally understand that since this is a more substantial change than we've been handling in our tickets so far there's more code to review and comment on and I bet most of my comments have to do with the way the code was originally written and not with what has been added.

Let me know if I can clarify anything

Model/DutyCalculator.php
lizzyWithReach: of course! totally understand, thank you!

Model/DutyCalculator.php
lizzyWithReach: Definitely! I appreciate how thorough you looked into this, it makes sense why we would need to save the quote every time and this background definitely helps. Do you think it would be helpful to fix this comment up with what you know about saving the quote / how the session and quote interact with the code as it exists right now?

My thinking is that I don't want to leave questions in the comments, I prefer to leave explanatory comments that give the reader more information that is directly relevant to the code around it. However, commenting styles are very individual/unique so I don't want to infringe on your personal style. Hopefully that comes across!

Model/DutyCalculator.php
lizzyWithReach: ohhh I totally understand now, I'll follow up on this in slack

Model/DutyCalculator.php
suraiyavic: Thanks Lizzy. This https://github.com/withreach/magento-plugin/pull/44/commits/43e8c0cb06c299c3f0cd941e2deecbe32d2c39de commit hopefully handles it.

Model/DutyCalculator.php
suraiyavic: https://github.com/withreach/magento-plugin/pull/44/commits/43e8c0cb06c299c3f0cd941e2deecbe32d2c39de should handle it.  
Remove the debug comment altogether though. Thanks.

Model/DutyCalculator.php
suraiyavic: Yeah. True. Thanks.

Model/DutyCalculator.php
suraiyavic: Miguel definitely can talk to DHL. 
If possible then let's talk about it tomorrow as well. Thanks!

Model/DutyCalculator.php
suraiyavic: Removed the question mark and updated the comment. https://github.com/withreach/magento-plugin/pull/44/commits/b4543f304265925b832162585e43be63bceff755 Thanks @lizzynibali

Model/DutyCalculator.php
suraiyavic: No ticket is necessary @lizzynibali. Magento is already tackling that. 
For example there are many types of response variables - even in dutyCalculator we have two types.
As for example \Reach\Payment\Api\Data\DutyResponseInterface that is also called response in the preexisting code (and I kept it the same) and then Magento\Framework\App\Response that is the HTTP response (if I am recalling it right). I added datatypes as part of @params to indicate which is which.
The response there is of  DutyResponseInterface category instead of the other one.
Would removing the  s from https or removing the comment altogether be more helpful to alleviate the confusion? Please let me know. I know I do not always express myself well enough. Thanks.

Model/DutyCalculator.php
suraiyavic: Shall be more than happy to discuss tomorrow @lizzynibali 
I did not know the meaning either - figured that out while playing with the code/tracing the code to fix MAG-70. We inherited this guy - it was there already. :D  
Thanks.

Model/DutyCalculator.php
suraiyavic: Adding more useful debug log (hopefully) for this section as I saw your other/newer statement by now. Thanks.

suraiyavic: This commit 87a79f49461544e2dea9c01befe0a2a40ff34b0c has different/updated log messages now. Please let me know if that is not clear enough/should be changed too. Thanks.

Model/DutyCalculator.php
suraiyavic: Commit 87a79f49461544e2dea9c01befe0a2a40ff34b0c would hopefully help. In the comment mentioned what type of response it is (immediately at that point the response is of type DutyResponseInterface). Removed the part that indicates what happens in general during transfer over internet. Thanks. (No new ticket is necessary).

Model/DutyCalculator.php
suraiyavic: Excellent point @lizzynibali . (As per the logic/flow) for country under such a situation (duty or shipment not allowed as per some admin config)  control flow is returning back immediately and no decision (unless I missed something due to still learning the system) is made there based on what is in saved quote (even if it is stale and has values specific to a prior country selection where duty is relevant). That said I do not know whether there is any report generation happening somewhere else in the system based on saved state (of quote) in the database or not. In that case I should cleanup saved state too. Thanks.

Model/DutyCalculator.php
suraiyavic: *Just copied over/reused this logic from pre-existing code. This is something from default admin config around DHL I think (which affects what behavior we see on store) which is one part of our conversation today as well. I have never seen that part fully working for me. The team indicated during yesterday's SU that I should be talking to you to have some clarity. Thanks.

Model/DutyCalculator.php
suraiyavic: I was wondering about something slightly different. I was wondering about if we are always rounding then is the pricing strategy (as set on admin panel like minimum, maximum) getting applied ever? May be I have to think/look around more. Thanks.

Model/DutyCalculator.php
suraiyavic: <h1>Handling different error messages:</h1>
I think it should be a new ticket. 
"I wonder if we should do different things based on the error received? " - I do not know as I do not know what the original requirement asked for as you know it is a balance between what the (original) requirement asked for and what is possible technically. This is the way pre-existing code was and I do not know whether I should assume the devs working on it followed the requirement or not. Is there any ticket listing original requirements when that initial implementation was done? (If it is on top of your head otherwise I can look on JIRA myself assuming that is the only ticketing system we have from the beginning). Or should I look in confluence?

I have to figure out some way to test it - either by designing a mock service imitating DHL and generate error or on purpose sending request that would return error or ... which obviously would increase development effort for this ticket. 

<h1>For another part which is important:</h1>
Additionally for this iteration I can set reachDuty (in session) to zero - this would beneeded when error is happening in the middle (country selection is changed and ...). If country selection for shipping address is not changed in between during the same session then checkbox would not appear due to the way the logic is (reachDuty is 0 or null at this stage ; checkbox appears when reachDuty is larger than 0).

**Oh wait (as I rechecked); equivalent effect is already achieved by setting duty to 0 in http response. So nothing to do for the section "For another part which is important:"**



All great points/questions overall. Thanks.

Model/DutyCalculator.php
lizzyWithReach: ok that makes sense, we can discuss this separately from this ticket :) thanks!

Model/DutyCalculator.php
lizzyWithReach: very good question, I thought that the pricing strategy changed the response from DHL's API and not the rounding mechanism for the value we get back from the API. But if you would like to test the different pricing strategies to see what it does that's not a bad idea

Model/DutyCalculator.php
suraiyavic: Done in commit: 87a79f49461544e2dea9c01befe0a2a40ff34b0c
Thanks.

Model/DutyCalculator.php
suraiyavic: Done - posted on MAG-70 on JIRA already (yesterday).

Model/DutyCalculator.php
suraiyavic: This piece is still outstanding - for what other countries, state/province name is relevant for accurate duty calculation.

suraiyavic: Okay so got the response from Miguel that for now state/province is relevant only for CA and BR as the countries of shipment. So current implementation done as part of MAG-70 is good in that area. 

Miguel  4:47 PM
Answer to the Canada / Brazil question - only those two at this time.
:heavy_check_mark:
1

Model/DutyCalculator.php
BronzePointKrel: Makes sense after reading everything Suraiya has put together regarding having to save into the session.

BronzePointKrel:
I've run this code locally and have tested through several iterations of currency, countries. It appears to be working locally. I did encounter a payment methods defect that only affects developers running a local version of the website. This is fixed in MAG-75. If running this ticket locally and you do not see Reach PayPal or Reach CC then that is fixed in MAG-75.

BronzePointKrel (approved):

Model/DutyCalculator.php
lizzyWithReach: If possible I would prefer the function name to be changed to be accurate (for ex. **fillOutResponseAndSessionOnError**) since nothing is changing with the Quote. 

Re-reading the function I don't think we need to set the quote since the response is coming back with an error, but if we do then we'll run into that issue when we get there. Any suggestions for resetting the quote would just be guessing and I don't want to recommend code changes off of a guess.

Let me know if that makes sense to you too

Model/DutyCalculator.php
lizzyWithReach: could you amend the comment in the code to clarify with the info we got from Miguel that for all other countries other than the ones in the special_countries list the state change doesn't matter? Just so that we don't have lingering q's that have been answered in the code

Model/DutyCalculator.php
```diff
@@ -116,84 +116,323 @@ public function __construct(
...
+            }
+        }
+    }
      /**
       * @inheritDoc
       */
     public function getDutyandTax($cartId, $shippingCharge, $shippingMethodCode, $shippingCarrierCode, $address, $apply = false)
     {
+        //this can go somewhere else more appropriate
+        $special_countries = array('CA','BR');
```
lizzyWithReach: instead of special_countries which is vague could this array be labeled something like countries_require_state to be more indicative of what this list means?

Model/DutyCalculator.php
suraiyavic: Pretty much tried to keep previous logic intact there (how the adjustment was done to the response returned by DHL). So this exploration can be done as a separate task possibly (I feel). Hey what do you feel about that? My reasoning: how many DHL calls are made that is different issue from whether pre-existing logic/calculation was honoring pricing strategy as per spec or not. So it may be a good idea not to add more time to this ticket and focus on what it is supposed to fix (so as not to cause scope creep). Thanks! @lizzynibali

Model/DutyCalculator.php
lizzyWithReach: Created ticket: https://withreach.atlassian.net/browse/MAG-76

Model/DutyCalculator.php
lizzyWithReach: decimals do show up, so I think I'm misreading this code. Going to close this out because I don't think we have any reason to dig into this further

Model/DutyCalculator.php
lizzyWithReach: created a ticket to check requirements: https://withreach.atlassian.net/browse/MAG-77

lizzyWithReach (changes requested):
Just a few requests for comment changes, variable name change, and function name change

Model/DutyCalculator.php
suraiyavic: Done as part of commit https://github.com/withreach/magento-plugin/pull/44/commits/b9d08fa382cc3a5934bca34cf7e1258d93a3b0f9
Thanks.

Model/DutyCalculator.php
suraiyavic: Changed the name as part of commit b9d08fa
Thanks.

lizzyWithReach: thank you!

lizzyWithReach (approved):

PR #46, opened 2020-02-02
[MAG-81] Generate a new DHL api call when there is a change in cart.

suraiyavic:
* Added new check so that state and country would not be considered as
unchanged when all of those are blanks.
* Added extra check while retrieving HSCode and country related information
from the database so that there is no error if nothing is retrieved from
the database.
* Added an observer/listener/handler that prepares the extension to generate
a new DHL call when there is a change in the shopping cart (due to adding or
 removing products).
* Associated cart change and the piece of code (handler) that does necessary
 cleanup needed for the system to trigger a new DHL api call.

* Have to add proper version based on when it goes out.


lizzyWithReach (approved):
This is great! Thank you

BronzePointKrel (approved):
This looks good. Your changes to CsvHsCode.php are the same that I did for my ticket that is also in code review :)

lizzyWithReach (approved):

PR #47, opened 2020-02-03
Mag 75 - Fix local vs public ip address lookup

BronzePointKrel:
This ticket only affects locally running Magento setups used during development. On these local systems, Reach payment methods (Paypal and CC) may not show up.

### DIFF
--- a/Model/Currency.php
+++ b/Model/Currency.php
@@ -142,6 +142,7 @@ class Currency extends \Magento\Framework\Model\AbstractModel
      */
     public function getLocalizeCurrency()
     {
+        $this->_logger->debug('------------- Currency::getLocalizeCurrency() : START ------------------');
         $rest = $this->httpRestFactory->create();
         $url = $this->reachHelper->getApiUrl();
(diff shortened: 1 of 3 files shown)


lizzyWithReach:
I have the same questions for this review as I left in the Varients review, so instead of duplicating comments here and potentially making things more confusing, I'll leave this comment here and address all concerns in the other review.

lizzyWithReach (changes requested):
I left a note in the conversation for this PR indicating that I have the same comments I would have left in this PR in the Varients PR. So we'll have our discussion there rather than having the same discussion in two places.

lizzyWithReach:
since we closed and deleted the other PR, my request for this PR would be to remove the test code / debug statements and leave the IP change as it is and we can merge that in. If that sounds ok with you @ryanjohnston

lizzyWithReach:
Not a pressing change, would need to spend time to double check removing the comments/debug statements and rather than spending the time to do that we're going to close this PR, close the ticket and revisit this when it becomes an issue again.

PR #48, opened 2020-02-03
Mag 78 - Add items with varients to the cart

BronzePointKrel:
This ticket allows users to add items with varients (i.e. sizes, colors, etc) to their cart. In the sample store, this includes shirts and pants.

Without this code fix, the duties and taxes checkbox does not show up if the cart has any items with varients.

### DIFF
--- a/Model/Currency.php
+++ b/Model/Currency.php
@@ -142,6 +142,7 @@ class Currency extends \Magento\Framework\Model\AbstractModel
      */
     public function getLocalizeCurrency()
     {
+        $this->_logger->debug('------------- Currency::getLocalizeCurrency() : START ------------------');
         $rest = $this->httpRestFactory->create();
         $url = $this->reachHelper->getApiUrl();
         $url = $url.'localize?MerchantId='.$this->reachHelper->getMerchantId();
@@ -149,10 +150,21 @@ class Currency extends \Magento\Framework\Model\AbstractModel
         if (!$this->checkLocalIP($ip)) {
             $url = $url.'&ConsumerIpAddress='.$ip;
         }
+
+        $this->_logger->debug('url: ');
+        $this->_logger->debug(json_encode($url));
+
         $rest->setUrl($url);
         $response = $rest->executeGet();
+
+        $this->_logger->debug('response: ');
+        $this->_logger->debug(json_encode($response));
+
         $result = $response->getResponseData();
         
+        $this->_logger->debug('result: ');
+        $this->_logger->debug(json_encode($result));
+
         if (isset($result['Currency'])) {
             return [
                 'currency'=>$result['Currency'],
@@ -160,6 +172,7 @@ class Currency extends \Magento\Framework\Model\AbstractModel
                 'country'=>$result['Country']
             ];
         }
+        $this->_logger->debug('------------- Currency::getLocalizeCurrency() : END ------------------');
         return false;
     }
 
@@ -222,6 +235,15 @@ class Currency extends \Magento\Framework\Model\AbstractModel
      */
     protected function checkLocalIP($ip)
     {
-        return in_array($ip, ['localhost','127.0.0.1','172.19.0.1','172.25.0.7','172.25.0.1']);
+        $this->_logger->debug('------------- Currency::checkLocalIP() : START ------------------');
+        if ( ! filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE) )
+        {
+            // is a local ip address
+            $this->_logger->debug('Using a local IP address');
+            return true;
+        }
+        $this->_logger->debug('Using a public IP address');
+        $this->_logger->debug('------------- Currency::checkLocalIP() : END ------------------');
+        return false;
     }
 }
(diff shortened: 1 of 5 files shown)


Model/ResourceModel/CsvHsCode.php
```diff
@@ -27,16 +27,28 @@ protected function _construct()
...
     public function getHsCode($sku)
     {
         $connection = $this->getConnection();
         $select = $connection->select()->from($this->getMainTable());
         $select->where('sku = ?', $sku);
         $result = $connection->fetchRow($select);
-        if (count($result) && isset($result['hs_code'])) {
-            return $result['hs_code'];
+
+        if ($result) {
```
lizzyWithReach, line 45: like you mentioned this is the same change that Suraiya made in the MAG-81 code review. Once we merge MAG-81 could you get that change into your branch and remove these changes so there's not a conflict? (I don't know if that's the right process but that's what I've got)

Model/ResourceModel/CsvHsCode.php
```diff
@@ -27,16 +27,28 @@ protected function _construct()
         $this->_init('reach_hs_code', 'id');
     }
 
+    public function testProductDetails($sku) {
```
lizzyWithReach, line 30: I'm a little worried about this testing code being sprinkled around the code base. I can see where you're going but it might make reading the code later more difficult since the test functions aren't in one place.

Did it help you when doing this code change to have these test methods? I'm good to keep them if I'm missing something

Model/Paypal.php
```diff
@@ -361,12 +361,46 @@ protected function getReferenceIdForRefund($payment)
      */
     public function isAvailable(\Magento\Quote\Api\Data\CartInterface $quote = null)
     {
+        $this->_logger->debug('---------------- isAvailable - START OF REQUEST----------------');
```
lizzyWithReach, line 364: I have the same concern with these debug statements, I don't know if it makes sense to leave these debug statements in the codebase. But please let me know if I'm missing something beneficial from leaving these debug statements in

Model/ResourceModel/CsvHsCode.php
lizzyWithReach: is this the fix to be able to add varient items to the cart? This is the only tangible code change I can find other than debug statements (of course I could 100% be blind)

Model/Currency.php
```diff
@@ -222,6 +235,15 @@ protected function getConsumerIp()
      */
     protected function checkLocalIP($ip)
     {
-        return in_array($ip, ['localhost','127.0.0.1','172.19.0.1','172.25.0.7','172.25.0.1']);
+        $this->_logger->debug('------------- Currency::checkLocalIP() : START ------------------');
+        if ( ! filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE) )
```
lizzyWithReach, line 239: I don't know where these constants FILTER_VALIDATE_IP, etc are defined. Is that code change missing or did I miss where these are defined?

BronzePointKrel: They are PHP constants defined by the run-time.

lizzyWithReach (changes requested):
my comments are a bit all over the place, I apologize in advanced if I missed the point with some of my comments.

I could use a bit of clarification and requested some cleanup to the code to make it easier to distinguish what changed to fix this issue.

Model/ResourceModel/CsvHsCode.php
BronzePointKrel: Yes, this fixes the varients.

BronzePointKrel: And yes, I can merge the two changes together.

Model/ResourceModel/CsvHsCode.php
BronzePointKrel: Even though the methods say 'test', they are for outputing values to the debug console. Because some classes do not allow the debugger to be imported, these functions pass these values back to where they can be outputed. I would recommend leaving them as they provide valuable debugging info and merchants can have the log level not set to 'debug'.

Model/Paypal.php
BronzePointKrel: These are paired with the 'test' functions to output valuable variables to the debug console for debugging.

Your choice on to leave them, comment them or remove them entirely.

BronzePointKrel:
> my comments are a bit all over the place, I apologize in advanced if I missed the point with some of my comments.
> 
> I could use a bit of clarification and requested some cleanup to the code to make it easier to distinguish what changed to fix this issue.

I've commented on your comments. Please let me know your thoughts.

Model/ResourceModel/CsvHsCode.php
suraiyavic: I have the exact same concerns as Lizzy has. In the spirit of good programming practices I guess it is more meaningful to change the names to something like getXYZ(..) . Thanks!

Model/ResourceModel/CsvHsCode.php
suraiyavic: I think putting extra check on `result` is already in MAG-81. If checking `result` is the only change in this branch in the exact same area/file then this branch can be deprecated altogether without any extra work I think . Is there anything else changed here that I am missing but would be needed for the fix of the issue with variant? Thanks!

BronzePointKrel: I think we can toss this entire branch. @lizzynibali ?

lizzyWithReach:
@ryanjohnston I'm with you on removing this branch to make things simpler. We can then take a look at the developer's change you made in the other PR to get that one merged in. I'll go ahead and close this PR/remove the branch if I can

PR #49, opened 2020-02-08
[MAG-79] Fixing Issues With Default Configuration

suraiyavic:
This work is fixing errors related to various default configurations
in pre-existing work which are as follows:
* Reference to non existent XPATHs
* Rreference to existing but wrong XPATHs
* Missing references to XPATHs required to show correct default
values.

It fixed problems in options array for dropdowns.

Notes: The original PrefTariffs option array is Ryan's work.
I adjusted it to make it working. Same is true for Pricing Strategy.
PHPStorm did some whitespace cleanup which made the work
look like having more changes than it really has. I am told to
 submit it as is.

See Also: MAG-66, MAG-65

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -19,8 +19,8 @@ class Data extends AbstractHelper
 
     const DHL_API_URL = 'https://api.dhlecommerce.com/';
     const DHL_SANDBOX_API_URL = 'https://api-sandbox.dhlecommerce.com/';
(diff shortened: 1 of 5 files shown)


suraiyavic:
MAG-85 is derived from this branch. So reviewing MAG-85 related pull request would cover both of these strongly related pieces in addition to MAG-66 and MAG-65. Thanks!

PR #50, opened 2020-02-08
Magento 85

suraiyavic:
So far handled only `Scope Type`. Depending on DHL use cases we may need to handle `Scope Code` and `Store ID`. Would know more about the requirements coming week. These are the reasons the PR is in draft mode.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -26,15 +26,15 @@ class Data extends AbstractHelper
     const DHL_DUTY_ALLOW_SPECIFIC_COUNTRY = 'reach/dhl/specificcountry';
     const DHL_DUTY_OPTIONAL_SPECIFIC = 'reach/dhl/optional_allowspecific';
     const DHL_DUTY_OPTIONAL_SPECIFIC_COUNTRY = 'reach/dhl/optional_specificcountry';
-    //would be needed for MAG-90 as well
-    const DHL_DUTY_ALLOW_SHIPPING = "payment/reach_payment/reach_dhl/applicable_shipping";
 
-    //***** xpath from system.xml as default does not exists *****/
-    const DHL_API_KEY = "payment/reach_payment/reach_dhl/dhl_key";
-    const DHL_API_SECRET = 'payment/reach_payment/api_secret';
-    const DHL_ITEM_SELLER = 'payment/reach_payment/reach_dhl/dhl_item_seller';
-    const DHL_PICKUP_ACCOUNT = "payment/reach_payment/reach_dhl/dhl_pickup_account";
-    const DHL_DEFAULT_HS_CODE = "payment/reach_payment/reach_dhl/default_hs_code";
+
+    //DHL_DUTY_ALLOW_SHIPPING would be needed for MAG-90 as well
+    const DHL_DUTY_ALLOW_SHIPPING = "reach/dhl/applicable_shipping";
+    const DHL_API_KEY = "reach/dhl/key";
+    const DHL_API_SECRET = "reach/dhl/api_secret";
+    const DHL_ITEM_SELLER = "reach/dhl/item_seller";
+    const DHL_PICKUP_ACCOUNT =  "reach/dhl/pickup_account";
+    const DHL_DEFAULT_HS_CODE =  "reach/dhl/default_hs_code";
 
     const CONFIG_REACH_ENABLED = 'reach/global/active';
     const CONFIG_CURRENCY_OPTION = 'reach/global/display_currency_switch';
@@ -44,8 +44,8 @@ class Data extends AbstractHelper
     const CONFIG_MERCHANT_ID = 'reach/global/mearchant_id';
     const CONFIG_API_SECRET = 'reach/global/api_secret';
 
-    //***** xpath from system.xml as default does not exists *****/
-    const CONFIG_CC_OPEN_CONCTRACT = 'payment/reach_payment/reach_cc/allow_open_contract';
+
+    const CONFIG_CC_OPEN_CONCTRACT = 'payment/reach_cc/allow_open_contract';
 
     const DHL_PREF_TARIFFS          = "reach/dhl/pref_tariffs";
     const DHL_PRICING_STRATEGY_PATH = "reach/dhl/pricing_strategy";
@@ -53,7 +53,9 @@ class Data extends AbstractHelper
     const DHL_END_USE_PATH          = "reach/dhl/end_use";
     const DHL_TRANSPORT_MODE_PATH   = "reach/dhl/transport_mode";
     const SANDBOX_MODE = 1;
-
+    //No getter for this two yet
+    const DHL_IMPORT_CSV_HS_CODE_PATH ="reach/dhl/import_csv_hs_code";
+    const DHL_EXPORT_CSV_HS_CODE_PATH ="reach/dhl/export_csv_hs_code";
     /**
      * Constant for payment
      */
@@ -121,8 +123,7 @@ class Data extends AbstractHelper
         if (isset($valueInWebsite)) {
             return  $valueInWebsite;
         }
-        $valueInDefault = $this->config->getValue($path,
-            \Magento\Framework\App\Config\ScopeConfigInterface::SCOPE_TYPE_DEFAULT);
+        $valueInDefault = $this->config->getValue($path,\Magento\Framework\App\Config\ScopeConfigInterface::SCOPE_TYPE_DEFAULT);
 
         return $valueInDefault; //At this point other higher priority scopes are undefined; if it is undefined too
                                 // then it would be treated as false
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -26,12 +26,15 @@ class Data extends AbstractHelper
     const DHL_DUTY_ALLOW_SPECIFIC_COUNTRY = 'reach/dhl/specificcountry';
     const DHL_DUTY_OPTIONAL_SPECIFIC = 'reach/dhl/optional_allowspecific';
     const DHL_DUTY_OPTIONAL_SPECIFIC_COUNTRY = 'reach/dhl/optional_specificcountry';
-    const DHL_DUTY_ALLOW_SHIPPING = 'reach/dhl/applicable_shipping';
-    const DHL_API_KEY = 'reach/dhl/key';
-    const DHL_API_SECRET = 'reach/dhl/api_secret';
-    const DHL_ITEM_SELLER = 'reach/dhl/item_seller';
-    const DHL_PICKUP_ACCOUNT = 'reach/dhl/pickup_account';
-    const DHL_DEFAULT_HS_CODE = 'reach/dhl/default_hs_code';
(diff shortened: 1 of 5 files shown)


Helper/Data.php
```diff
@@ -84,28 +102,36 @@ public function getStashApiUrl()
...
-    public function getConfigValue($field, $storeId = null)
+    public function getConfigValue($path, $storeID = null, $scopeID = null)
     {
-        return $this->scopeConfig->getValue(
-            $field,
-            ScopeInterface::SCOPE_STORE,
-            $storeId
-        );
+        $valueInStore = $this->config->getValue($path, \Magento\Store\Model\ScopeInterface::SCOPE_STORE);
+        if (isset($valueInStore)) {
```
suraiyavic, line 125: This is the main change (checks `Scope Type`) in this commit.
It checks a lower priority `Scope Type` only if the higher priority `Scope Type`s are all undefined/null.

suraiyavic:
MAG-79, MAG-66 and MAG-65 are getting handled as part of this ticket. Thanks.

Model/Config/Source/PrefTariffs.php
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+     * Return array of options as value-label pairs, eg. value => label
+     *
+     * @return array
+     */
+    public function toOptionArray()
+    {
+        return [
+            ['value' => 1, 'label' => __('True')],
```
lizzyWithReach, line 16: since the values are still 1 and 0 won't 1/0 be sent through to DHL? It's good that the labels are set to True/False but my understanding of MAG-65 was that True/False needed to be sent along to DHL in the FLC API request.

This might help: https://magento.stackexchange.com/questions/139148/magento-2-how-to-read-the-values-in-etc-config-xml

Using the code above,

```
$value = $this->scopeConfig->getValue('section/group/value1', \Magento\Store\Model\ScopeInterface::SCOPE_STORE);  

will return 1 and

$value = $this->scopeConfig->isSetFlag('section/group/value1', \Magento\Store\Model\ScopeInterface::SCOPE_STORE);  

will return true.
```

Model/Config/Source/PricingStrategy.php
```diff
@@ -13,8 +13,8 @@ class PricingStrategy implements \Magento\Framework\Option\ArrayInterface
     public function toOptionArray()
     {
         return [
-            ['value' => 'MINIMUM', 'label' => __('MINIMUM')],
             ['value' => 'MAXIMUM', 'label' => __('MAXIMUM')],
+            ['value' => 'MINIMUM', 'label' => __('MINIMUM')],
```
lizzyWithReach, line 17: Wanted to confirm my assumptions:
- this re-organized for aesthetic purposes
- The default (MAXIMUM) is set in the etc/config.xml file

thanks!

Model/DutyCalculator.php
```diff
@@ -655,11 +650,11 @@ protected function getQuote($request, $accessToken)
         // $this->_logger->debug('----------------GET QUOTE FROM DHL - START OF REQUEST----------------');
         // $this->_logger->debug(json_encode($url));
         // $this->_logger->debug(json_encode($accessToken));
-        // $this->_logger->debug(json_encode($request));
+        $this->_logger->debug(json_encode($request));
         // $this->_logger->debug('================GET QUOTE FROM DHL - END OF REQUEST================');
```
lizzyWithReach: is it worthwhile to delete these other commented debug statements? I see the value in the debug statements that you've left in uncommented.

Model/DutyCalculator.php
```diff
@@ -585,24 +586,18 @@ protected function prepareRequest($freightCharge, $shippingAddress)
     protected function getShippingOrigin()
     {
         $origin = [];
-        $region = $this->_scopeConfig->getValue(
-            self::XML_PATH_ORIGIN_REGION_ID,
-            \Magento\Store\Model\ScopeInterface::SCOPE_STORE,
-            $this->storeManager->getStore()->getId()
-        );
 
+        $region = $this->reachHelper->getShippingOriginState(self::XML_PATH_ORIGIN_REGION_ID,
```
lizzyWithReach, line 590: would you happen to know why the XML_PATH_ORIGIN_COUNTRY_ID and XML_PATH_ORIGIN_COUNTRY_ID constants are defined in this file rather than with the rest of the paths in the Helper/Data.php file? 

I'm wondering if it would be cleaner to move these paths to the Data.php file the way the rest of the configuration details are so that these calls can go from:

```
$this->reachHelper->getShippingOriginCountry(
            self::XML_PATH_ORIGIN_COUNTRY_ID, $this->storeManager->getStore()->getId());
```

to

`$this->reachHelper->getShippingOriginCountry();`

and line up with the other config getter functions. What do you think?

Model/Config/Source/PrefTariffs.php
suraiyavic: Actually there is a piece of code in pre-existing DutyCalculator.php that turns this 1/0 to true, false and as far as DHL is concerned it would get right value. 
DHL call would fail with full commit by Ryan as he changed both config and DutyCalculator.php in a way that DHL would get string `True` or `False` with his committed code as is.

Model/Config/Source/PricingStrategy.php
suraiyavic: No, not for aesthetic purpose. When both key and value are string ... if you do not use the default one as the first item it would not show you correct one in the beginning without  extra manipulation.

Model/DutyCalculator.php
suraiyavic: About the commented out debug level log statements - those are from Ryan's and I can definitely remove. I un-commented the one that I used. The uncommented one would not cause any problem on production as there is a magento command  to turn this on or off. On production it would be off. We had a lengthy discussion on this back in January on Slack. Recommended approach is leaving logging statements as is (instead of commenting those out) and turning on and off those based on which env it is - dev/qa/prod. This is true and supported in most of the programming languages.

Model/DutyCalculator.php
suraiyavic: Original developer chose to do it the constant could be easily shifted to Data.php. I left the constant/it there as is -  just shifted the getter logic to Data.php for proper handling of that variable with respect to scope (default, website, store). Just changed the minimum that I must need to change/move.

suraiyavic: For sure I can shift those to data.php and can make it similar to the other getters ... would need another extra round of (env setup and) testing at my end though

suraiyavic: May be instead of doing it now we can do that on that batch of tickets (with a small ticket for this one) when we are pushing dynamic scoping of CC and PayPal (on hold and explained at the bottom of MAG-93)?

Model/Config/Source/PricingStrategy.php
lizzyWithReach: thank you! I'm happy I checked.

Model/DutyCalculator.php
lizzyWithReach: It's certainly not required for this review to be completed. So if you would like to move those constants, awesome! But if not, or you would like to save that work for later that's totally ok too.

Model/DutyCalculator.php
lizzyWithReach: awesome, if you get to it before the review is closed then great! If not, that's ok. It's a small thing.

Helper/Data.php
```diff
@@ -19,54 +21,82 @@ class Data extends AbstractHelper
...
 
-    const CONFIG_CC_OPEN_ONCTRACT = 'payment/reach_cc/allow_open_contract';
 
-    const DHL_PREF_TARIFFS          = "payment/reach_payment/reach_dhl/pref_tariffs";
-    const DHL_PRICING_STRATEGY_PATH = "payment/reach_payment/reach_dhl/pricing_strategy";
-    const DHL_CLEARANCE_MODE_PATH   = "payment/reach_payment/reach_dhl/clearance_mode";
-    const DHL_END_USE_PATH          = "payment/reach_payment/reach_dhl/end_use";
-    const DHL_TRANSPORT_MODE_PATH   = "payment/reach_payment/reach_dhl/transport_mode";
-    const SANDBOX_MODE = 1;
+    const CONFIG_CC_OPEN_CONCTRACT = 'payment/reach_cc/allow_open_contract';
```
lizzyWithReach: could you correct this const variable? From CONFIG_CC_OPEN_CONCTRACT to CONFIG_CC_OPEN_CONTRACT

Helper/Data.php
```diff
@@ -228,7 +293,7 @@ public function canAllowMultiPleCurrency()
      */
     public function allowCurrencySpeicifcCountry()
```
lizzyWithReach, line 295: you could comment this spelling error also the way you did for getDisaplyBade, if that would be helpful for easily finding these typos later

lizzyWithReach (changes requested):
The only code change that I would prefer to have done before approving this pull request would be the typo "CONFIG_CC_OPEN_CONCTRACT" to be fixed since that typo was introduced by this code change. 

All other comments were either to learn more about the code changes or make suggestions but I don't see them as required to close this review.  Thank you!

Helper/Data.php
suraiyavic: Sure thing. And sorry ... my mistake while adding missing C on the pre-existing `CONFIG_CC_OPEN_ONCTRACT`  (on previous line 43 above) I did not pay attention that there were other spelling mistakes there.

lizzyWithReach: all good! there are a lot of nitpicky things going on here, I can totally see how that would happen.

Helper/Data.php
suraiyavic: Just added a comment indicating the spelling error in the latest commit https://github.com/withreach/magento-plugin/commit/7839c5d4d81a2bda6427cd0af37885e7b62b9a48 . Thanks.

Helper/Data.php
suraiyavic: > The only code change that I would prefer to have done before approving this pull request would be the typo "CONFIG_CC_OPEN_CONCTRACT" to be fixed since that typo was introduced by this code change.
> 
> All other comments were either to learn more about the code changes or make suggestions but I don't see them as required to close this review. Thank you!

Removed commented out debug log from Ryan and corrected the spelling error.
Also added a note (as suggested) indicating a place to correct spelling error in future. All of these are part of commit https://github.com/withreach/magento-plugin/commit/7839c5d4d81a2bda6427cd0af37885e7b62b9a48

 About the other comment: 
The small task of making all the getter methods looking the same by pushing the constants (XML_PATH_ORIGIN_COUNTRY_ID, XML_PATH_ORIGIN_REGION_ID) as defined in the DutyCalculator.php out to Data.php would need adjustment in the code at four different places. Shall handle that as a small ticket when we tackle the batch of tickets to fix pre-existing CC and PayPal code.
Thanks!

lizzyWithReach (approved):
Great! Thanks Suraiya

PR #51, opened 2020-02-24
Magento 39 & CC in MAG-93

suraiyavic:


Model/Cc.php
```diff
@@ -217,6 +217,10 @@ public function __construct(
         $this->errorHandler = $errorHandler;
     }
 
+    /** Why do we need both isAvailable and isActive
```
lizzyWithReach, line 220: so this is where the /getPaymentMethods API call would help. We need to know that Paypal, for example, is available for this customer and check that it's enabled for this magento instance.

So we do need both, but the way that it is written right now is not correct for credit cards. We would need to check at the specific card level, not just credit cards in general.  We'll have to talk through this to outline the details before changing the way this code is functioning right now.

composer.json
```diff
@@ -2,7 +2,7 @@
   "name": "reach/reachpayment",
   "description": "",
   "type": "magento2-module",
-  "version": "2.3.5",
+  "version": "2.3.6",
```
lizzyWithReach, line 5: do you think these changes and the MAG-85 changes will go out together? I would assume, just thinking that if they do go out together then we would only need the version number bumped in this code review

lizzyWithReach (approved):
No code changes necessary from what I can tell, but I'll leave this review open until we get MAG-85 wrapped up since the MAG-85 code changes are in this PR as well.

Model/Cc.php
suraiyavic: Yes agree. Thanks!

Model/DutyCalculator.php
```diff
@@ -560,7 +559,7 @@ protected function prepareRequest($freightCharge, $shippingAddress)
                     $itemData['hsCode']=$this->reachHelper->getDhlDefaultHsCode();
                 }
                 $itemData['skuNumber']=$item->getSku();
-                $itemData['itemValue']=['value'=>$item->getRowTotal()/$item->getQty(),'currency'=>$quote->getQuoteCurrencyCode()];
+                $itemData['itemValue']=['value'=>($item->getRowTotal() - $item->getDiscountAmount())/$item->getQty(),'currency'=>$quote->getQuoteCurrencyCode()];
```
suraiyavic, line 562: This was the main change here

composer.json
suraiyavic: Good point. Both MAG-39 and MAG-85 can go out together. If not then MAG-85 should be released first. That said how about considering Adam's view too as he will be testing and both being together may make his life difficult. Let me talk and then adjust the version for mag-85 and mag-39 accordingly. Does that sound good? Thanks!

lizzyWithReach (approved):
All good! Agreed to push the MAG-93 CC fixes as well with this commit w/the team

PR #53, opened 2020-03-09
Magento 101

suraiyavic:


suraiyavic:
There are lot of places (in the pre-existing code) with code duplication instead of leveraging a helper/util file (for code reuse).  That can be refactored as part of a separate ticket.

Model/Currency.php
```diff
@@ -22,8 +22,9 @@ class Currency extends \Magento\Framework\Model\AbstractModel
     /**
      *  @var \Psr\Log\LoggerInterface
      */
-    protected $_logger; 
+    protected $_logger;
 
+    const PRECISION = 2;
```
lizzyWithReach, line 27: Created this ticket: https://withreach.atlassian.net/browse/MAG-109 to address my concerns round this value being hardcoded. Totally understand why it's hardcoded at this point in time.

Model/Currency.php
```diff
@@ -212,16 +213,37 @@ protected function getConsumerIp()
...
-     * is regarding Line 71 in Model/Reach.php.
-     *
+     * check IP is local machine IP
+     * @param  string $ip
      * @return boolean
      */
     protected function checkLocalIP($ip)
     {
-        return in_array($ip, ['localhost','127.0.0.1','172.19.0.1','172.25.0.7','172.25.0.1']);
+        //Ryan's code from MAG-75 (to be able to test payment code properly)
```
lizzyWithReach: this comment was definitely helpful for me, thank you! Could you remove it before we push the code? I don't think this would benefit anyone other than me as a reviewer when looking at this code.

Model/Cc.php
```diff
@@ -495,9 +498,12 @@ protected function _buildCheckoutRequest($payment, $amount)
         }
         $request['Items']=[];
         foreach ($order->getAllVisibleItems() as $item) {
+            if ($item->getProductType() == "simple" && ($item->getParentItem())) {
```
lizzyWithReach, line 514: I think this is odd enough logic that a comment would help us remember why we are doing this for the future.  For example, "For variant items, the item gets added to the cart twice. Once without configuration details and one with configuration settings. This logic skips the duplicate item information without configuration details"

Also, this is a handy way for me to double check that how I'm reading this is accurate haha so please correct me if I'm wrong.

Model/PayPalManagement.php
```diff
@@ -198,14 +198,17 @@ protected function _buildCheckoutRequest($payment,$deviceFingerprint)
         $request['ReferenceId'] = $order->getIncrementId();
         $request['ConsumerCurrency']= $order->getOrderCurrencyCode();
         $request['DeviceFingerprint'] = $deviceFingerprint;
-        $order->getOrderCurrencyCode();
+        $order->getOrderCurrencyCode(); //what it is for?
```
lizzyWithReach: this looks like nonsense, did you try commenting it out? The same call is used a couple of lines above to set the ConsumerCurrency which makes sense, but this call looks superfluous

lizzyWithReach: you know what, I wonder if originally this call was set to a variable that could be used in all of the convertCurrency calls below and they just took out the variable part?

Model/Paypal.php
```diff
@@ -421,7 +421,7 @@ protected function _buildCheckoutRequest($payment, $amount)
         $request['MerchantId'] = $this->reachHelper->getMerchantId();
         $request['ReferenceId'] = $order->getIncrementId();
         $request['ConsumerCurrency']= $order->getOrderCurrencyCode();
-        $order->getOrderCurrencyCode();
+        $order->getOrderCurrencyCode(); //What it is for?
```
lizzyWithReach, line 424: Could you explain the difference between PayPalManagement and Paypal? They look like they're doing the same thing and I'm not sure why we need two different files for Paypal

Model/ReachConfigProvider.php
```diff
@@ -130,9 +130,16 @@ protected function getConsumerIp()
      */
     protected function checkLocalIP($ip)
```
lizzyWithReach, line 131: can we put this checkLocalIP in the helper file rather than here and in Currency.php? if it's outside of the scope of this ticket that's ok

Model/ReachCurrency.php
```diff
@@ -106,7 +106,7 @@ public function getCurrencyRates($currency, $toCurrencies = null)
     {
         if (in_array($this->state->getAreaCode(), ['frontend','webapi_rest']) && $this->reachHelper->isReachCurrencyEnabled()) {
             if (!$this->_storeManager->getStore()->isCountryApplicable()) {
-                return parent::getRate($toCurrency);
+                return parent::getRate($toCurrency); //where is this $toCurrency coming from???
```
lizzyWithReach, line 109: I'm honestly confused about what this entire section is doing, $toCurrency would be null and I think it's a typo from a copy/paste from the other functions that have $toCurrency as the second parameter since this function has $toCurrencies instead of $toCurrency

lizzyWithReach: I recognize that this is overriding the getCurrencyRates method from the Magento Currency parent class but I don't know why. I don't know if these comments are helpful haha but I'm thinking out loud

lizzyWithReach (changes requested):
I have some q's and clarifications, the only tangible change I'm requesting is removing one comment (that exists in two places) and adding a new comment (to code that exists in 3 places).  These aren't high intensity changes so it's not a blocker, but I won't close the review until you get a chance to review the comments to make sure my questions don't just get lost (if that's ok!) Thank you!!

Model/Currency.php
suraiyavic: It is hard coded in the spirit of leveraging your suggested shortcut in MAG-100 :
"**There is a quick solution we could get to (always rounding to 2 decimal places)** or the absolute correct way to do it (always round to the number of units that the selected currency supports, but no more than 2 digits due to a bug in our API)"

I can definitely take the longer route of setting precision to 0/1/2 if that is coming back from our API call and set anything (coming back from API) that is above 2 to 2 if that is what we want or cannot live without.  Thanks Lizzy!

Model/Currency.php
suraiyavic: I can definitely remove the comment - only thing is I wanted to make sure that I do not get credit for someone else's work. Thank you.

Model/PayPalManagement.php
suraiyavic: Honestly I am not sure why it is that way. I wanted to add the comment by side to start a conversation on it during review.

Model/Paypal.php
suraiyavic: Hey Lizzy, 
Short answer is I do not know. In January when I started here I was asking the same question and I still have it. Honestly I believe the the code should be  refactored. During dealing with Mag-93 the discussion outcome was PayPal code is half done and should be dealt with as part of separate ticket. Thanks!

Model/ReachConfigProvider.php
suraiyavic: It should be in a helper file as it is in minimum two different files in our code base. 
But i think that type of code duplication cleanup can be handled as a separate ticket.

Model/Cc.php
suraiyavic: Sure thing. I shall add the commit # once it is up in github. At present why the logic was used is explained only in the JIRA ticket.

Model/ReachCurrency.php
suraiyavic: I flagged this/added the comment - so that it is looked at or get fixed at some point.

Model/Currency.php
suraiyavic: Commit 2cd3c3ed80ed1342fd4446d9bb06386d5b7b4138
Thanks.

Model/Currency.php
lizzyWithReach: yeah! This is perfect for right now, just wanted to record the extra work of making precision dynamic for the future. I'd rather get this change out and come back to not having this value be hardcoded than slow things down by getting that work done now if that makes sense.

Model/Paypal.php
lizzyWithReach: Ok no problem, thank you!

Model/ReachConfigProvider.php
lizzyWithReach: https://withreach.atlassian.net/browse/MAG-110 created a ticket, thanks!

lizzyWithReach (approved):
Thanks Suraiya!!

Model/Cc.php
suraiyavic: Commit 451f0032ee87472a3926180be34681c9f0f054e4
Thanks.

Model/PayPalManagement.php
suraiyavic: It feels like a typo to me. Removing it (unless there is an objection). 
Commit: ffe3f6442e04e765e9356cdc9e60f93815dffa0a
Thanks.

lizzyWithReach (approved):

lizzyWithReach:
Not merging this PR because the code changes included here were merged with the MAG-102 PR.

PR #54, opened 2020-03-10
[MAG-104] Changed Config Path for Credit Card Sort Order

suraiyavic:
Changed the `Config Path` that stores where on the UI the Credit Card info
appears with respect to other UI components.

Should be part of release 2.3.6


lizzyWithReach (approved):
Looks good! Simple change

PR #55, opened 2020-03-12
Magento 102

suraiyavic:
Second half of the problem (Core Magento bug affecting our scheduled job) is still there to be taken care of.  This is the reason behind the DRAFT PR.

### DIFF
--- a/Model/Currency.php
+++ b/Model/Currency.php
@@ -219,7 +219,6 @@ class Currency extends \Magento\Framework\Model\AbstractModel
      */
     protected function checkLocalIP($ip)
     {
-        //Ryan's code from MAG-75 (to be able to test payment code properly)
         if (!filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE))
         {
             // is a local ip address
--- a/Model/Currency.php
+++ b/Model/Currency.php
@@ -114,6 +114,7 @@ class Currency extends \Magento\Framework\Model\AbstractModel
             foreach ($rates as $rate) {
                 $data = $this->getResource()->getByCurrency($rate['Currency']);
                 if (count($data) && isset($data[0]['rate_id'])) {
+                    //update
                     $this->setData([
                         'rate_id'=>$data[0]['rate_id'],
                         'offer_id'=>$rate['Id'],
@@ -121,7 +122,10 @@ class Currency extends \Magento\Framework\Model\AbstractModel
                         'rate'=>$rate['Rate'],
                         'expire_at'=>$rate['Expiry']
                     ])->save();
+                    $this->storedData = []; //this is one work around for this Magento bug
+                    // (https://github.com/magento/magento2/issues/4174) as explained in JIRA MAG-102
                 } else {
+                    //insert
                     $this->setData([
                         'rate_id'=>null,
                         'offer_id'=>$rate['Id'],
--- a/Model/Currency.php
+++ b/Model/Currency.php
@@ -22,8 +22,9 @@ class Currency extends \Magento\Framework\Model\AbstractModel
     /**
      *  @var \Psr\Log\LoggerInterface
      */
-    protected $_logger; 
+    protected $_logger;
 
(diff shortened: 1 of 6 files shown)


Model/Currency.php
```diff
@@ -113,14 +114,18 @@ public function updateRates()
...
                 if (count($data) && isset($data[0]['rate_id'])) {
+                    //update
                     $this->setData([
                         'rate_id'=>$data[0]['rate_id'],
                         'offer_id'=>$rate['Id'],
                         'currency'=>$rate['Currency'],
                         'rate'=>$rate['Rate'],
                         'expire_at'=>$rate['Expiry']
                     ])->save();
+                    $this->storedData = []; //this is one work around for this Magento bug
```
lizzyWithReach, line 125: just wanted to confirm my understanding, this fixes the Magento bug because the bug was that updates to the database included old information in the storedData array. By clearing this array pre-emptively you're making sure that the next update doesn't have old data getting stored in the database and only valid, up to date information is making it into the database. Is this accurate?

lizzyWithReach (approved):
Approving this review, I just had one question to get clarification. I'll leave this review open until we can get MAG-101 approved and merged, then merge MAG-100 then this review if that sounds ok?

Model/Currency.php
suraiyavic: Let's say during update scenario (not insert) Magento would like to determine which column of a row has new value. To determine that - previous state of a row should be compared to incoming column values for that exact same row.  Due to a Magento bug (as reported by others on net as well) the comparison was happening between two different rows (with respect to the incoming values for both) instead of previous and new values for the exact same row. 

As a hypothetical example if 
-  In the incoming data two rows have updated value in the `expired_at field  
And
-  The new values in that `expired_at` is same for both rows in the incoming data

Then the `expired_at` was getting updated only for the very first row with this new expired_at value and any rows coming later in the loop with the exact same `expired_at` were incorrectly not getting updated in the database due to that faulty comparison/logic. 

The idea was to clear everything in the storedData associative array (that was holding information about a previous row the code just looked instead of the previous state of the current row). There is a non public function in Magento to do so/reset()/clear which is not available from our extension but is available via $this->storedData (another craziness) . 

 Please let me know if more elaborate explanation is required. It was equally confusing/did not make sense situation for me when I was trying to locate root cause of this problem and had to go several level deep in the call chain to find a logic which seems okay on the surface but it is not due to the way it is comparing one row to another instead of stored values of one row with the incoming /new value for the same row. I completely understand/relate to the confusion and would be more than happy to explain in different manner/style.


Due to a Magento bug the code was `comparing` the field values for `two different rows` (related to two different currencies) to decide which fields of current row under consideration have been changed. The correct logic would be to compare previous state of the fields (from database) versus incoming values of the fields (from API call) for `the exact same row` (currency).

suraiyavic:
> Approving this review, I just had one question to get clarification. I'll leave this review open until we can get MAG-101 approved and merged, then merge MAG-100 then this review if that sounds ok?

Thanks Lizzy! That sounds good. Responded to all of your comments. Shall tackle the actual file changes  when I am recharged a bit to make sure I am not introducing silly mistakes.

Model/Currency.php
lizzyWithReach: makes more sense now, thank you!

lizzyWithReach (approved):

Model/Paypal.php
```diff
@@ -421,7 +421,7 @@ protected function _buildCheckoutRequest($payment, $amount)
         $request['MerchantId'] = $this->reachHelper->getMerchantId();
         $request['ReferenceId'] = $order->getIncrementId();
         $request['ConsumerCurrency']= $order->getOrderCurrencyCode();
-        $order->getOrderCurrencyCode();
+        $order->getOrderCurrencyCode(); //What it is for?
```
lizzyWithReach: do we also want to remove the redundant call here?

suraiyavic: Just did it Lizzy 7784476775073aa1e7c6532a4c616f046ea40967
Thanks.

lizzyWithReach (approved):

PR #56, opened 2020-03-20
[MAG-112] Fix for Unspecified Option in the Clearance Mode Drop down

suraiyavic:
Need to discuss who would be released before whom - this one or the batch of mag-100, mag-101, mag- 102 or all three of those along with this ticket.
Would set the merging or cherry picking (whatever is appropriate accordingly).
Thanks!

### DIFF
--- a/Model/Config/Source/ClearanceMode.php
+++ b/Model/Config/Source/ClearanceMode.php
@@ -12,10 +12,18 @@ class ClearanceMode implements \Magento\Framework\Option\ArrayInterface
      */
     public function toOptionArray()
     {
+        /*We are setting value to 'N/A' instead of '' or ' '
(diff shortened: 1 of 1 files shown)


suraiyavic:
This is a fix for issue found on tag 2.3.6.
The fix was done from branch that came out of code commit with tag 2.3.6.

If we want to merge it to current master then the approach would be:

1) git clone https://github.com/withreach/magento-plugin . 

2) git checkout magento_112

git branch (checking where we are at)

3) git checkout master
git branch (checking where we are at)


4) git merge magento_112

( source magento_112
destination master)
112 is getting merged to master

Check where we are at:
git branch

To push it to master

5) git push origin master

lizzyWithReach (approved):
I love the extra detail in the comment, I know it adds to the line count but in this instance it's super helpful. Thank you!

PR #57, opened 2020-03-25
Magento 107

suraiyavic:
Fix for Open Contract ID and Currency not being Saved properly

### DIFF
--- a/Observer/SalesOrderPlaceAfterObserver.php
+++ b/Observer/SalesOrderPlaceAfterObserver.php
@@ -129,11 +129,11 @@ class SalesOrderPlaceAfterObserver implements ObserverInterface
 
                 if ($response_extracted  && isset($response_extracted['Payment'])) {
                     $detail=[];
-                    $detail['currency']= $response_extracted ['ConsumerCurrency'];
-                    $detail['method']=$response_extracted ['Payment']['Method'];
-                    $detail['identifier']=$response_extracted ['Payment']['AccountIdentifier'];
+                    $detail['currency']= $response_extracted['ConsumerCurrency'];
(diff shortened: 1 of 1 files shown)


Observer/SalesOrderPlaceAfterObserver.php
```diff
@@ -117,15 +119,22 @@ protected function fetchContractDetail($contractId)
...
-                    $detail['method']=$response['Payment']['Method'];
-                    $detail['identifier']=$response['Payment']['AccountIdentifier'];
+                    $detail['currency']= $response_extracted['ConsumerCurrency'];
+                    $detail['method']=$response_extracted['Payment']['Method'];
+                    $detail['identifier']=$response_extracted['Payment']['AccountIdentifier'];
 
-                    if (isset($response['Times']['Expiry'])) {
-                        $expredAt = explode('T', $response['Times']['Expiry']);
+                    if (isset($response_extracted['Times']['Expiry'])) {
+                        $expredAt = explode('T',$response_extracted['Times']['Expiry']);
```
lizzyWithReach: can you fix the expredAt typo here to "expiredAt"?

lizzyWithReach (changes requested):
Just a little typo, if you think it's reasonable to fix that at this time Suraiya. (if it's actually a weed of a typo that exists everywhere then I'm ok leaving it)

Observer/SalesOrderPlaceAfterObserver.php
suraiyavic: Sure Lizzy. Here you go: 298f55e308cde2677e9a1d46ae7ccc0926cd6562

lizzyWithReach (approved):
Thank you!

PR #58, opened 2020-03-26
Magento 117

suraiyavic:
Would like to test some edge cases and different work flows before setting it to regular PR.

### DIFF
--- a/Model/DutyCalculator.php
+++ b/Model/DutyCalculator.php
@@ -275,7 +275,6 @@ class DutyCalculator implements \Reach\Payment\Api\DutyCalculatorInterface
         //or whether applying duty is a must for that country or not (a setting in magento admin
         //panel)
         //filling out both Quote and session state as appropriate
-        //DHL quoteID is already saved into session; so not doing it here again
         $quote = $this->checkoutSession->getQuote();
         $duty_adjusted = $this->priceCurrency->round($duty); //copied over pre-existing code
         //but are we supposed to round always?
--- a/Model/DutyCalculator.php
+++ b/Model/DutyCalculator.php
@@ -148,6 +148,7 @@ class DutyCalculator implements \Reach\Payment\Api\DutyCalculatorInterface
                " parameter that is passed to this function or because duty is not optional for the chosen country ".
                "(shipment address)");
            $quote->setReachDuty($this->checkoutSession->getReachDuty());
+           $quote->setBaseReachDuty($this->checkoutSession->getBaseReachDuty());
            $quote->setDhlQuoteId($this->checkoutSession->getDhlQuoteId());
         }
         else {
@@ -155,6 +156,7 @@ class DutyCalculator implements \Reach\Payment\Api\DutyCalculatorInterface
             //as apply duty and tax(DT) is not selected or it is optional; duty and tax would not be used in total
             // pricing; so setting all relevant quote values to zero
             $quote->setReachDuty(0);
+            $quote->setBaseReachDuty(0);
         }
         $quote->save();//this is needed so that different aspects related to a quote are available on other
         // pages/propagate to other pages (without changing existing code on those pages).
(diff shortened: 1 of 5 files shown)


Model/DutyCalculator.php
```diff
@@ -288,7 +290,6 @@ public function fillOutQuoteAndSessionUsingFeeReturned($duty, $address, $apply,
             $this->checkoutSession->setBaseReachDuty(0);
             $this->checkoutSession->setReachDuty($duty_adjusted);
             $quote->setDhlQuoteId('');
-            $this->checkoutSession->setDhlQuoteId('');
```
lizzyWithReach, line 291: if the session isn't getting cleared when the checkbox isn't selected, then couldn't a customer 

- go through the checkout flow
- click the duties and taxes checkbox (the session gets a quoteId), 
- leave the checkout flow then when they 
- go through checkout again they 
- don't click the checkbox/deselect checkbox, 
- wouldn't their session still have the old quoteId?

Just want to double check that removing the clearing of the quoteId in the session doesn't result in issues like the one I outlined above

Model/DutyCalculator.php
```diff
@@ -278,6 +279,7 @@ public function fillOutQuoteAndSessionUsingFeeReturned($duty, $address, $apply,
             $quote->setDhlBreakdown(json_encode($response['feeTotals']));
             $this->checkoutSession->setDhlBreakdown($quote->getDhlBreakdown());
             $this->checkoutSession->setApply(true); //checkbox selection /corresponding passed value
+            $this->checkoutSession->setDhlQuoteId($response['quoteId']);
```
lizzyWithReach: The comment on line 260 says: `//DHL quoteID is already saved into session; so not doing it here again` , if we want to save the quote into the session again (which I don't mind, it should be the same value and if not then this is a good chance to correct that) could you delete this comment if it makes sense to do so?

lizzyWithReach (changes requested):
I could use a bit of clarification and I have a comment removal suggestion, thank you!

Model/DutyCalculator.php
suraiyavic: Few things Lizzy:

Some values are not getting cleared from session (deliberately). This is to be able to reuse the D&T (as well as DHLQuoteid) values without a new DHL call (if nothing else changes except the checkbox selection ).  
What is getting into the database is from Quote not from Session.
So when user is deselecting the apply D&T checkbox - the info related to D&T is getting cleared from the Quote but not from Session. This is so that just in case the user select the apply D&T option again we can reload value from session to quote without making a new DHL call. If we clear the DHLQuoteID from session then if the user changes mind during the same session then we would not have the DHLquoteid without making a new call.
However there are cases when it would be a problem (like as listed in MAG-115). This is the reason I was requesting (please have a look at `dev-magento` slack channel ) last Friday to bump up the priority of MAG-115 so that more event handlers (like MAG-81) are added for proper handling of such cases. However that did not happen (I am going to share the link with you shortly). Thanks!


 @lizzynibali

Model/DutyCalculator.php
suraiyavic: Hello Lizzy, Shall have a look/cleanup once my system is not `in transition` state. Right now updating few things on my dev setup. Thanks!

Model/DutyCalculator.php
lizzyWithReach: ohh thank you! I appreciate the time you spent clarifying this

Model/DutyCalculator.php
lizzyWithReach: sounds good! take your time

suraiyavic:
```
//DHL quoteID is already saved into session; so not doing it here again
        $quote = $this->checkoutSession->getQuote();
```
I can remove the comment. Thanks!

suraiyavic: commit 15153627c4dedfebbb8417849bfb680a56b6e1f5 
Thank you. 
On  merging this back to magento-25 branch (as that is cut from this magento-117 branch)

lizzyWithReach: yay! thanks!

lizzyWithReach (approved):

PR #59, opened 2020-03-27
Magento 25

suraiyavic:


Model/DutyCalculator.php
```diff
@@ -225,7 +231,6 @@ public function handleCaseWhereDutyIsAlreadyRetrieved($apply, $quote, $address)
             $this->handleTaxApplicability($address, $apply);
             $this->_logger->debug('User wants to apply the D&T. Value of apply ::' . $apply);
             $quote->setBaseReachDuty($this->checkoutSession->getBaseReachDuty());
-            $quote->setReachDuty($this->checkoutSession->getReachDuty());
```
lizzyWithReach, line 228: could you clarify why this line is being removed, but the `$quote->setBaseReachDuty($this->checkoutSession->getBaseReachDuty());` wasn't? I'm confused what the benefit is of removing these set calls

Model/DutyCalculator.php
```diff
@@ -244,6 +249,20 @@ public function handleCaseWhereDutyIsAlreadyRetrieved($apply, $quote, $address)
         $this->response->setDuty($this->checkoutSession->getReachDuty());
     }
 
+    /** gets duty in base currency from display currency
+     * @param float $duty
+     * returns float
+     */
+    public function getBaseDuty($duty)
+    {
+        //it may look at stale value???
```
lizzyWithReach: could you explain a bit more about what this comment means? I don't have enough context to know if something should be done about this, if this is a bug, etc

Model/DutyCalculator.php
```diff
@@ -261,36 +280,35 @@ public function fillOutQuoteAndSessionUsingFeeReturned($duty, $address, $apply,
...
-        //should not that be based on corresponding admin setting?
+        //should not that be based on corresponding admin setting? or what is coming back from REACH API
         $this->checkoutSession->setReachDuty($duty_adjusted);
 
+        $baseDuty = $this->getBaseDuty($duty_adjusted);
+
+        $this->checkoutSession->setBaseReachDuty($baseDuty);
+
+        $this->checkoutSession->setDhlQuoteId($response['quoteId']);
+        //$this->checkoutSession->setReachDuty($duty_adjusted);
```
lizzyWithReach: this comment can be removed since this same call is being made on line 284 (just a few lines above) `$this->checkoutSession->setReachDuty($duty_adjusted);`

Model/DutyCalculator.php
```diff
@@ -261,36 +280,35 @@ public function fillOutQuoteAndSessionUsingFeeReturned($duty, $address, $apply,
...
         if ($apply || !$this->getIsOptional($address->getCountryId())) {
             //checkbox selection was duty should be applied
             //or applying duty is a must for that country
-            $baseCurrency = $this->storeManager->getStore()->getBaseCurrency();
-            $rate = $baseCurrency->getRate($baseCurrency->getCode());
-            $baseDuty = $duty_adjusted / $rate;
             $quote->setBaseReachDuty($baseDuty);
-            $this->checkoutSession->setBaseReachDuty($baseDuty);
             $quote->setReachDuty($duty_adjusted);
-            $this->checkoutSession->setReachDuty($duty_adjusted);
```
lizzyWithReach, line 276: with the removal of these session calls only a part of the session is being updated in this function that is named: `fillOutQuoteAndSessionUsingFeeReturned` could you help explain why these lines were removed? and potentially rename the function to better reflect what is actually happening here?

Let me know if I'm just wildly missing the point haha thank you!

lizzyWithReach (changes requested):
I have questions about consistency, I couldn't track the code clearly enough to work out why some lines were removed and others were added. I didn't find a lot of information in the ticket, so it's entirely possible that I'm missing vital context. Thank you for your patience with me!

Model/DutyCalculator.php
suraiyavic: Yeah, sure. 
$quote->setReachDuty($this->checkoutSession->getReachDuty()); is removed because it was already done in the very first line of this handleCaseWhereDutyIsAlreadyRetrieved(....) definition. The line that was removed is a duplicate/ redundant line.

Model/DutyCalculator.php
suraiyavic: The note was pretty much like talking to myself/set as a hint that there might be a window of uncertainty /invalid rate based on when rate was fetched and when it is applied to get base currency equivalent from display currency. Not sure what would be the frequency of such occurrences and corresponding implications. I think you mentioned this briefly when you were demonstrating the payment section in front of us (why fetching every six hour is reasonable ... and so on)  Lizzy. Thanks! Do you want me to remove this comment?

Model/DutyCalculator.php
suraiyavic: Right. Thanks! Just handled this. (Shall add the commit hash once I  push the code.)

Model/DutyCalculator.php
lizzyWithReach: ohh that makes sense, I could see either: 

- expanding the comment to include more of what you were thinking about
- removing the comment

it's a small thing, so only if it makes sense to you too

Model/DutyCalculator.php
suraiyavic: Actually the statements are taken out and placed outside that conditional block not entirely removed. This is because those are needed for both (if and else) cases if we want to use session as temporary memory and copied back to Quote model to save those value in the database (when the `apply D&T`  is pressed/forced without giving a the user a choice). If `apply d&t` checkbox is deselected then the copy from session to quote model does not happen though the D&T data is available in session if needed. The function is named as `fillOutQuoteAndSessionUsingFeeReturned` as this function fills out quote and session data structures with fees returned from DHL API call at some point. Do you have any suggestions for a more meaningful name? Thanks.

Model/DutyCalculator.php
suraiyavic: Removed. Thanks! commit 971e7cc32fde044bdeb84a06f3d5308baf9713cb

Model/DutyCalculator.php
suraiyavic: commit 971e7cc32fde044bdeb84a06f3d5308baf9713cb 
Thanks!

Model/DutyCalculator.php
lizzyWithReach: thanks for the explanation! No this is all good, I needed to look at the function overall.

Model/DutyCalculator.php
lizzyWithReach: thank you!

lizzyWithReach (approved):

PR #60, opened 2020-03-31
Magento 109

suraiyavic:


Helper/Cacher.php
```diff
@@ -0,0 +1,110 @@
...
+use Magento\Framework\App\Cache\Type\FrontendPool;
+use Magento\Framework\Cache\Frontend\Decorator\TagScope;
+
+use Magento\Framework\Serialize\SerializerInterface;
+
+
+/**
+ * Helper Class
+ */
+//class Cacher extends AbstractHelper
```
suraiyavic: Forgot to clean this. Shall do so in next commit.

Helper/Cacher.php
```diff
@@ -0,0 +1,110 @@
...
+            }
+            $this->data = $expandedData;
+            $expandedData = $this->serializer->serialize($expandedData);
+            $this->_logger->debug("Saving Currency Precision in Cache :::" . $expandedData);
+            $this->cache->save($expandedData, self::REACH_CACHE_ID, array(self::REACH_CACHE_TAG));
+        }
+    }
+
+    /*
+     * To do; not required now
```
suraiyavic, line 120: Can remove it too.

Helper/Cacher.php
```diff
@@ -0,0 +1,110 @@
...
+     *  @var \Psr\Log\LoggerInterface
+     */
+    protected $_logger;
+
+    /**
+     * @var SerializerInterface
+     */
+    private $serializer;
+
+    const REACH_CACHE_ID = 'reach_payment_custom_cache';
```
suraiyavic, line 42: Needed for saving and retrieving data.

Helper/Cacher.php
```diff
@@ -0,0 +1,110 @@
...
+    const REACH_CACHE_TAG = 'REACH_PAYMENT_CUSTOM_CACHE';
```
suraiyavic, line 46: Is needed for clearing or flushing a specific cache instead of clearing everything (or something with wider scope).

etc/cache.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Cache/etc/cache.xsd">
```
suraiyavic, line 7: This file is needed to show this custom cache option on Admin panel. When it is visible on admin and enabled a merchant/admin with enough privilege can flush or clean it from there.

Setup/UpgradeSchema.php
```diff
@@ -15,6 +15,30 @@ public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $con
         $installer->startSetup();
         $connection = $installer->getConnection();
 
+        $precisionTable = $installer->getTable('reach_currency_precision');
```
suraiyavic, line 24: New table in the database to store currency versus precision information.

Model/ResourceModel/Currency.php
```diff
@@ -36,8 +36,39 @@ public function getByCurrency($code)
         return $connection->fetchAll($select);
     }
 
+    //should write  deletion routine too
+
+    /**
+     * @param $currencyCode string
+     * @return array
+     */
+    public function getPrecisionByCurrency($currencyCode)
```
suraiyavic, line 47: To retrieve precision info from the database. I think this comment should reside in the code itself.

Model/ResourceModel/Currency.php
```diff
@@ -36,8 +36,39 @@ public function getByCurrency($code)
...
+            ->where('currency_code = ?', $currencyCode);
+        return $connection->fetchAssoc($select);
+    }
+
+    /**
+     * @param $currencyCode string
+     * @param $precision integer
+     * @return \Zend_Db_Statement_Interface
+     */
+    public function setPrecisionByCurrency($currencyCode, $precision)
```
suraiyavic, line 71: To save precision info into the database. I think this comment should reside in the code itself.

Helper/Cacher.php
```diff
@@ -0,0 +1,110 @@
...
+use Magento\Framework\Cache\Frontend\Decorator\TagScope;
+
+use Magento\Framework\Serialize\SerializerInterface;
+
+
+/**
+ * Helper Class
+ */
+//class Cacher extends AbstractHelper
+class Cacher extends TagScope
```
suraiyavic, line 29: If this class is not a subclass of `TagScope` then we get error when we try to compile code for dependency injection and interceptor generator as well as while running cache cleanup or flushing options.

Model/Currency.php
```diff
@@ -235,18 +245,60 @@ protected function checkLocalIP($ip)
...
+                $currencyVsPrecision = [$currencyCode => $precision];
+                $this->cacher->saveDataInCache($currencyVsPrecision);
+            }
+            $precision_adjusted = ($precision > self::PRECISION_CUTOFF)? self::PRECISION_CUTOFF: $precision; //because of a bug in
+            //our API (as per the note in description of JIRA MAG-100)
+            $this->_logger->debug("in convert currency: precision adjusted :".$precision_adjusted);
+
         }
-        return round($amount, self::PRECISION);
+        return round($amount, $precision_adjusted); //when nothing coming back from api or db; verify with team
```
lizzyWithReach, line 299: ideally in the unlikely event that you don't get anything back from the db, cache or Reach API I can see two approaches:

- default to 0 precision because that will work with all currencies, and the merchant/us will likely notice it faster
- default to 2 precision because that matches what most currencies need anyway, but anything with 0 precision (JPY) will fail across the board

lizzyWithReach: the safer choice that retains customer's ability to checkout would be defaulting to 0, so I'm leaning that way

Helper/Cacher.php
lizzyWithReach: would it make sense to leave this note as a comment in the code? I wouldn't have guessed that a caching helper class would need to extend something called "TagScope"

Helper/Cacher.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+namespace Reach\Payment\Helper;
+
+//use Magento\Framework\App\Helper\AbstractHelper;
```
lizzyWithReach: general question, does this cache exist on the server? or per session? just to confirm my understanding

Helper/Cacher.php
lizzyWithReach: it makes sense to me to have a way to invalidate this cache, so I'm ok leaving this stubbed code here. You're right, we don't necessarily need it right now but worst case we delete the commented code in a year or so if it turns out we never needed to implement it

lizzyWithReach (changes requested):
as always your code is super clean and readable, it was so easy to follow what was happening here. Thank you!

I'm leaving the review open to handle setting a default precision if nothing is in the cache, db and the Reach API request fails and also you mentioned removing some commented code in a future commit.

I didn't see any bugs in the code itself although I wonder if there's a way to get the cache into an invalid state.

Model/Currency.php
lizzyWithReach: this is how the code is written currently, it defaults to 0 precision

Helper/Cacher.php
suraiyavic: Hello Lizzy @lizzynibali ,
 The cache exists on the server - not on client end.
Also the scope is way beyond a user session. And here is my reasoning 
"[MAG-109] Added Custom Magento Cache to Store Precision

Most appropriate place to save precision values related to various currencies
is cache not session. This is because the precision of a currency does not change
 from one customer to another. ..."

Please let me know if I am not articulating or some info is missing.
Thanks!

lizzyWithReach: yes ok, I love it. Just wanted to double check that I wasn't fundamentally misunderstanding anything. Thank you!

Helper/Cacher.php
suraiyavic: Done. Commit fcd15604f707f0a182d8bfbdf276f33dd0097cca 

Thanks Lizzy!  @lizzynibali

Model/ResourceModel/Currency.php
```diff
@@ -36,8 +36,51 @@ public function getByCurrency($code)
...
+
+     * @param $currencyCode string
+     * @param $precision integer
+     * @return \Zend_Db_Statement_Interface
+     */
+    public function setPrecisionByCurrency($currencyCode, $precision)
+    {
+        $connection = $this->getConnection();
+
+        $connection->insert('reach_currency_precision', [
```
lizzyWithReach, line 75: would it be worthwhile to handle the case when insert doesn't actually update anything? (if 0 is returned by the insert() vs what we expected (1))

lizzyWithReach (changes requested):
Had a question on what we should do if insert() doesn't return what we expect, thanks for your helpful comments Suraiya!

Model/ResourceModel/Currency.php
lizzyWithReach: discussed with Suraiya, she considered throwing an exception in this case but to avoid breaking the site because of a thrown exception we're going to leave this code as is.

If the precision details can't be added to the database for some reason, it should still be cached. If it can't be cached either then another /localize call will be made. If a thousand /localize calls are hitting Reach then we'd notice right away.

It would be an edge edge case for the database to be inaccessible.

lizzyWithReach (approved):
Thank you for your patience with me!

PR #61, opened 2020-04-16
[MAG-118] Turning Default HS Code Field into a Required one with Some Data…

suraiyavic:
Finished the ticket as per current understanding.
There are some clarifications/confirmations needed which are posted in the JIRA ticket.

### DIFF
--- a/view/adminhtml/web/js/validation.js
+++ b/view/adminhtml/web/js/validation.js
@@ -1,0 +1,25 @@
+require([
+        'jquery',
+        'mage/translate',
+        'jquery/validate'],
+    function($){
+        $.validator.addMethod(
+            "reach-default-hs-code-validator", function (hscode) {
+                //Got help/hints from
+                //https://magento.stackexchange.com/questions/95171/magento-2-form-validation
+                //https://magento.stackexchange.com/questions/139023/custom-field-validation-in-system-xml
+                //https://regex101.com/
+                //allows twenty characters (after removing `.` and `-`) in the hscode string.
+                // With `.` and/or `-` the string length may be longer than 20 from this routine
+                //as of now; however that is getting checked and restricted again by built in max length checker:
+                //`validate-length maximum-length-20`.
+                //At the end we are good
+                //Does not accept two consecutive special characters but allows mix and match of `.` and `-` in the
+                //hscode string
+                //check is done only at frontend level
+                //additionally could be done/enforced when getting saved in db (in a path way bypassing UI)
+
+                return (/^[\d]{2}([\.\-]?[A-Za-z0-9]){0,18}$/.test(hscode));
+            }, $.mage.__('Please enter a string that starts with two digits and have at least 2 characters and at most 20 characters (letter and numbers)'));
+    }
+);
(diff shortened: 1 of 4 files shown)


view/adminhtml/web/js/validation.js
```diff
@@ -0,0 +1,22 @@
...
+                //https://magento.stackexchange.com/questions/95171/magento-2-form-validation
+                //https://magento.stackexchange.com/questions/139023/custom-field-validation-in-system-xml
+                //https://regex101.com/
+                //This routine is enforcing good enough check on length so we possibly can remove the built in
+                // validation of the form `validate-length minimum-length-2 maximum-length-20` from system.xml.
+                //Does accept two consecutive special characters. Also allows mix and match of `.` and `-` in the
+                //hscode string - made it that way after talking (on Slack) to the team.
+                //Check is done only at frontend level
+                //Additionally the check could be done/enforced when getting saved in db (in a path way bypassing UI)
+                hscode = hscode.trimRight();
```
lizzyWithReach: do we care about any whitespace at the beginning of the hscode? It would be silly but a user could do "  12.123.B12  "

view/adminhtml/web/js/validation.js
```diff
@@ -0,0 +1,22 @@
...
+                return (/^[\d]{2}[A-Za-z0-9\.-]{0,18}$/.test(hscode));
+            }, $.mage.__('Please enter a string that starts with two digits and have at least 2 characters and at most 20 characters (letter and numbers)'));
```
lizzyWithReach: if this is the error message if this validation doesn't work, could explicitly say "Please enter an HS Code that starts..." because that's what the field is? Or would you prefer this error message be more vague ("a string")

lizzyWithReach (approved):
I had a couple of comments, but this code looks good so no need to stop it from going out. If you have the time to look at my comments that would be great, thank you!

view/adminhtml/web/js/validation.js
suraiyavic: I would love to put a better error message that you suggested. The message is something I came up with but it sounds more developer oriented than end user oriented (I agree)   😄 Thanks Lizzy!

view/adminhtml/web/js/validation.js
suraiyavic: Yeah I was wondering should I take the requirement statement of it starts with two number way too literally or not.  With that hesitation in mind I took it too literally and trimmed space only from the right hand side. It is super easy to trim from both ends and I am more than happy to do so Lizzy. Thanks!

lizzyWithReach: yay!

view/adminhtml/web/js/validation.js
lizzyWithReach: haha for sure! Thank you :)

suraiyavic:
(After discussion with the internal team) leveraged `Before Plugin` type of approach (one of the prescribed/preferred coding style in Magento) for data cleanup.

view/adminhtml/web/js/validation.js
suraiyavic: Commit 4d5773a1339a054f289c08c348e8d9d7b9e4aacf (in addition to addressing your concerns used something similar to the second approach described below)

Thanks Lizzy!  @lizzynibali

lizzyWithReach (approved):
Yay! Thanks Suraiya, looks good

PR #63, opened 2020-04-21
Magento 123

suraiyavic:


etc/adminhtml/system.xml
```diff
@@ -183,6 +183,13 @@
                         <validate>required-entry validate-length minimum-length-2 maximum-length-20 validate-no-empty reach-default-hs-code-validator</validate>
                         <config_path>reach/dhl/default_hs_code</config_path>
                     </field>
+                    <!-- Default Country of Origin of a Product - related to Default HS Code -->
+                    <field id="default_country_of_origin" translate="label" type="multiselect" sortOrder="115" showInDefault="1" showInWebsite="1" showInStore="0">
```
lizzyWithReach: I would have expected the default country of origin to have the same visibility as default_hs_code, but showInWebsite is 0 for default country of origin but 1 for default_hs_code. Wanted to check that that's ok

lizzyWithReach (changes requested):
Just want to check on the visibility of the default country of origin

etc/adminhtml/system.xml
lizzyWithReach: sorry I meant "showInStore", not showInWebsite

suraiyavic: commit 2f694bcc469db49bf3846ec11d7005abb7edb0ac
Thanks!

lizzyWithReach (approved):
Thank you!

PR #65, opened 2020-04-22
Magento 86

suraiyavic:


Plugin/Magento/Checkout/Block/Checkout/LayoutProcessor.php
```diff
@@ -0,0 +1,23 @@
...
+     * Process js Layout of block.
+     * UI item is getting removed
+     * @param \Magento\Checkout\Block\Checkout\LayoutProcessor $subject
+     * @param \Closure $proceed
+     * @param array $jsLayout
+     * @return array
+     */
+    public function aroundProcess($subject, $proceed, $jsLayout)
+    {
+        unset($jsLayout['components']['checkout']['children']['sidebar']['children']['summary']['children']['totals']['children']['grand-total']['config']['basicCurrencyMessage']);
```
lizzyWithReach, line 20: this is crazy haha thank you for the references in the ticket, that was very helpful. Do you think it would make sense to link to Magento official documentation about this?

although maybe people who work with M2 are used to this, or future devs could just look up the ticket..

lizzyWithReach (approved):
another comment q, approving because it's not a big enough point to stop this from going out. Thank you!

Plugin/Magento/Checkout/Block/Checkout/LayoutProcessor.php
suraiyavic: commit 7773de5ad3c6d1d0c4894180cf3b1d2f9e007757
Thanks!

lizzyWithReach (approved):
Perfect, thank you!

PR #67, opened 2020-10-20
[MAG-156] Fix so that PayPal appears as an Option

suraiyavic:


dpwithreach (approved):
Looks like this should fix it, but I'm curious about why the `\Reach\Payment\Helper\Data` defined twice as two different variables (reachHelper and paymentData)

image

suraiyavic:
> Looks like this should fix it, but I'm curious about why the `\Reach\Payment\Helper\Data` defined twice as two different variables (reachHelper and paymentData)
> 
> image

Hey Daniel, If you have a closer/careful look then you would see that one is : Reach/Payment/Helper/Data and the other is Magento/Payment/Helper/Data . We need both. First one is part of our plugin and the second one is coming from core Magento. Thanks!

dpwithreach:
oops🤦‍♂️
Clearly I overlooked that 👍

PR #69, opened 2021-04-05
D-180 Merchant Id is blank - fix

RotateAt60MPH:
Typo in one config key path.  It was corrected to the proper spelling at one time, but the old (incorrect) spelling should have been kept.  Descoped fixing it in the DB since it would have involved a DB migration step as part of the upgrade.


brenleydueck (approved):

lizzyWithReach:
We haven't seen an issue with `merchant_id` when working on this locally, I don't believe this change is required but if it is we can bring this PR back.

PR #70, opened 2021-12-29
EXT-644: Error handling check for Message key

grandel-withreach:
- When an error is provided by the Checkout API, the Plugin assumes that the `Message` key will always be present. This is not the case.
- Added a conditional check to first determine if the `Message` key is present before attempting to use it.


lizzyWithReach (approved):
Thank you Grandel! I'm happy you were able to confirm the fix, as simple as it is it's always nice to see it working as expected on the site itself.

marcwenger (approved):

PR #73, opened 2022-06-16
MAGENTO-17: Implement webapi route for Notification

grandel-withreach:
- Created a new webapi route for handling the Reach Notifications
- Needed override the `DeserializerFactory` in order to be able to handle the `x-www-form-url-encoded` request that is coming from the Checkout API
- Replaced the `Notify` URL with the webapi route endpoint
- Removed the `Callback` Controller

### DIFF
--- a/Model/Cc.php
+++ b/Model/Cc.php
@@ -90,6 +90,12 @@ class Cc extends \Magento\Payment\Model\Method\Cc
      */
     protected $_clientTimeout = 45;
 
+    /**
+     * Payment Method feature
+     *
+     * @var bool
+     */
+    protected $_isGateway = true;
 
     /**
      * Payment Method feature
@@ -595,10 +601,7 @@ class Cc extends \Magento\Payment\Model\Method\Cc
      */
     private function getCallbackUrl($order)
     {
-        $url = $this->coreUrl->getUrl('reach/cc/callback', [
-            '_secure' => true,
-            '_store'  => $order->getStoreId()
-        ]);
+        $url = $this->coreUrl->getUrl('rest/default/V1/reach/notification');
         $url .= "?orderid=" . $order->getQuoteId();
         return  $url;
     }
--- a/Model/Cc.php
+++ b/Model/Cc.php
@@ -482,7 +482,7 @@ class Cc extends \Magento\Payment\Model\Method\Cc
         $request['Consumer'] = $consumer;
         
         
-        $request['Notify'] = $this->getCallbackUrl($order);
+        $request['Notify'] = $this->getNotifyUrl($order);
         $request['ConsumerCurrency']= $order->getOrderCurrencyCode();
 
         $rateOfferId =  $this->reachCurrency->getOfferId($order->getOrderCurrencyCode());
@@ -599,7 +599,7 @@ class Cc extends \Magento\Payment\Model\Method\Cc
      * @param object $order
      * @return string
      */
-    private function getCallbackUrl($order)
+    private function getNotifyUrl($order)
     {
         $url = $this->coreUrl->getUrl('rest/default/V1/reach/notification');
         $url .= "?orderid=" . $order->getQuoteId();
--- a/Model/ReachNotification.php
+++ b/Model/ReachNotification.php
@@ -56,7 +56,12 @@ class ReachNotification implements NotificationInterface
             if ($this->validate($request, $signature)) {
                 // TODO:: Handle other Notification order states
                 $response = json_decode($request, true);
-                if (isset($response['OrderState']) && $response['OrderState'] == "Processed" && $response['UnderReview'] === false && !count($response['Refunds'])) {
+
+                $isProcessed = isset($response['OrderState']) && $response['OrderState'] == "Processed";
+                $isNotUnderReview = $response['UnderReview'] === false;
+                $hasNoRefunds = !count($response['Refunds']);
+
+                if ($isProcessed && $isNotUnderReview && $hasNoRefunds) {
                     $order = $this->loadOrder($response['ReferenceId']);
                     $order = $this->orderRepository->get($order->getId());
                     $order->getPayment()->accept();
--- a/Model/ReachNotification.php
+++ b/Model/ReachNotification.php
@@ -1,58 +1,81 @@
 <?php
 
-namespace Reach\Payment\Controller\Cc;
+namespace Reach\Payment\Model;
 
+use Reach\Payment\Api\Data\ResponseInterface;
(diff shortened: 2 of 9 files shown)


etc/di.xml
```diff
@@ -12,8 +12,38 @@
...
 
+    <preference for="Reach\Payment\Api\NotificationInterface" type="Reach\Payment\Model\ReachNotification" />
+
     <preference for="Reach\Payment\Api\Data\StashResponseInterface" type="Reach\Payment\Model\Api\Data\StashResponse" />
     <preference for="Reach\Payment\Api\Data\DutyResponseInterface" type="Reach\Payment\Model\Api\Data\DutyResponse" />
     <preference for="Reach\Payment\Api\Data\ResponseInterface" type="Reach\Payment\Model\Api\Data\Response" />
     <preference for="Reach\Payment\Api\Data\HttpResponseInterface" type="Reach\Payment\Model\Api\Data\HttpResponse" />
+    <preference for="Reach\Payment\Api\Data\NotificationResponseInterface" type="Reach\Payment\Model\Api\Data\NotificationResponse" />
+
+    <type name="Magento\Framework\Webapi\Rest\Request\DeserializerFactory">
```
grandel-withreach, line 23: This is overriding the default Magento `DeserializerFactory`. Needed to add an `item` for all the other types as well (`application/json`, `application/xml`, etc) using the default Magento classes as well, otherwise it wouldn't know how to handle them.

Webapi/Rest/Request/Deserializer/XWwwFormUrlencoded.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Reach\Payment\Webapi\Rest\Request\Deserializer;
+
+
+use Magento\Framework\App\State;
+use Magento\Framework\Phrase;
+use Reach\Payment\Model\PostbackNotification\Decoder;
+
+class XWwwFormUrlencoded implements \Magento\Framework\Webapi\Rest\Request\DeserializerInterface
```
grandel-withreach, line 10: A lot of this content follows the default `Deserializer` classes native to Magento. Reference: https://github.com/pepe1518/magento2/blob/master/vendor/magento/framework/Webapi/Rest/Request/Deserializer/Json.php

Model/Cc.php
```diff
@@ -601,10 +601,7 @@ protected function getConsumerInfo($order)
      */
     private function getCallbackUrl($order)
```
lizzyWithReach: would it be a hassle to also rename this getter? 

or is this an expected function name that we shouldn't change? 
(ex expected by Magento to look like this, in which case should we adjust our terminology from notification to callback in `rest/default/V1/reach/notification`?)

Model/ReachNotification.php
```diff
@@ -0,0 +1,98 @@
...
+     * @param $signature
+     * @return ResponseInterface
+     */
+    public function handleNotification($request, $signature)
+    {
+        try {
+            if ($this->validate($request, $signature)) {
+                // TODO:: Handle other Notification order states
+                $response = json_decode($request, true);
+                if (isset($response['OrderState']) && $response['OrderState'] == "Processed" && $response['UnderReview'] === false && !count($response['Refunds'])) {
```
lizzyWithReach: could we make this line more readable? it's a little long and also we could wait to do this until we get a linter/checkstyle in place

Model/ReachNotification.php
```diff
@@ -0,0 +1,98 @@
...
+        $this->_orderFactory = $orderFactory;
+        $this->orderRepository = $orderRepository;
+    }
+
+    /**
+     * @param $request
+     * @param $signature
+     * @return ResponseInterface
+     */
+    public function handleNotification($request, $signature)
```
lizzyWithReach, line 53: is it magento magic that takes the notification, uses webapi.xml to find this method then split the body into the request and signature parts for what this method is expecting as parameters?

lizzyWithReach (approved):
one question and two nitpicks, thanks for the comments Grandel! helped a ton!

Model/Cc.php
grandel-withreach: This method is only ever called in this class so we could change it to whatever we like

I also tested it locally and confirmed that changing it doesn't have any negative impacts. 

What do you think of `getNotifyUrl`? Or would `getNotificationUrl` be more fitting?

Model/Cc.php
```diff
@@ -599,12 +599,9 @@ protected function getConsumerInfo($order)
      * @param object $order
      * @return string
      */
-    private function getCallbackUrl($order)
+    private function getNotifyUrl($order)
```
lizzyWithReach, line 602: oh this is great! thanks! I was good with either Notify or Notification especially since it's only used in this class

Model/ReachNotification.php
```diff
@@ -0,0 +1,103 @@
...
+        try {
+            if ($this->validate($request, $signature)) {
+                // TODO:: Handle other Notification order states
+                $response = json_decode($request, true);
+
+                $isProcessed = isset($response['OrderState']) && $response['OrderState'] == "Processed";
+                $isNotUnderReview = $response['UnderReview'] === false;
+                $hasNoRefunds = !count($response['Refunds']);
+
+                if ($isProcessed && $isNotUnderReview && $hasNoRefunds) {
```
lizzyWithReach, line 64: oh I find this much clearer, thank you!!

lizzyWithReach (approved):

Model/ReachNotification.php
grandel-withreach: Yeah pretty much, the configuration defined in the `webapi.xml` for this route what should be ran,

```
<route url="/V1/reach/notification" method="POST">
        <service class="Reach\Payment\Api\NotificationInterface" method="handleNotification"/>
        <resources>
            <resource ref="anonymous" />
        </resources>
    </route>
```

So looking at this route configuration, we've defined that for this endpoint I want it to use the `NotificationInterface` as the Service and call the `handleNotification` method to handle the request. 

This `ReachNotification` class implements the `NotificationInterface` which defines that this class should have a `handleNotification` class that receives the two parameters. To my understanding, its Magento magic take looks at the incoming requests and checks to see if the request body contains the `request` and `signature` makes it available through those params. 

The request must also match the parameters defined in the `NotificationInterface`. DId a quick test and changes `$request` to `$test` and it failed, because my request body did not contain `test`

lizzyWithReach: love it! thank you for confirming! Especially when there's a bit of magic it helps to know my assumptions line up with what appears to be happening.

PR #74, opened 2022-06-16
[MAGENTO-37] write down basic deployment steps from dev to qa to prod

lizzyWithReach:
So I don't need to remember all the details next time we're ready to push to QA or prod

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ If you are interested in using this plugin please contact sales@withreach.com
 4. merge to QA branch
 5. delete feature branch
 6. deploy to QA environment
-   1. Access QA EC2 Server
+   1. Access QA Server
    2. run `bin/composer install`
    3. which will pull the qa branch code into the magento site with recent updates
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,3 +2,19 @@
...
 If you are interested in using this plugin please contact sales@withreach.com
+
+# Deployment Pipeline
+1. work on dev feature branch
+2. open PR, merging to QA branch
+3. PR approved
+4. merge to QA branch
+5. delete feature branch
+6. deploy to QA environment
+   1. Access QA EC2 Server
```
grandel-withreach: Considering this is a public repo do we want to expose that were using a EC2 server?

grandel-withreach (approved):

README.md
lizzyWithReach: well, they can see everything including this PR haha and you're right. Updated.

I wonder if we can update it so that only certain branches are public?

PR #83, opened 2022-07-15
Non base refund failure fix

ChantalDelachevrotiere:
Refund request was sending in base currency and not consumer currency causing `RefundLimitExceeded` in some instances. 

Fix: grab `getGrandTotal()` from the credit memo which is the grand total of the refund from the credit memo in the purchase currency (consumer currency).

### DIFF
--- a/Model/Cc.php
+++ b/Model/Cc.php
@@ -59,7 +59,7 @@ class Cc extends \Magento\Payment\Model\Method\Cc
      *
      * @var bool
      */
-    protected $_canRefund = false;
+    protected $_canRefund = true;
 
     /**
      * Availability option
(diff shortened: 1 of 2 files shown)


Model/Cc.php
```diff
@@ -418,7 +418,7 @@ public function refund(\Magento\Payment\Model\InfoInterface $payment, $amount)
         $request=[];
         $request['OrderId'] = str_replace('-capture', '', $payment->getParentTransactionId());
         $request['MerchantId']= $this->reachHelper->getMerchantId();
-        $request['Amount']= $amount;
+        $request['Amount']= $this->reachCurrency->convertCurrency($payment->getOrder()->getOrderCurrencyCode(), $payment->getCreditmemo()->getGrandTotal());
```
grandel-withreach: Could we make this a bit more readable by assigned some of the values into variables?

```
$currencyCode = $payment->getOrder()->getOrderCurrencyCode();
$grandTotal = $payment->getCreditmemo()->getGrandTotal();
$request['Amount']= $this->reachCurrency->convertCurrency($currencyCode, $grandTotal);
```

Same comment would be applicable to the changes in `Model/Paypal.php`

grandel-withreach (approved):

PR #87, opened 2022-07-22
return generic auth fail for fraud related rejections for CC and Paypal

ChantalDelachevrotiere:
When a transaction is rejected by Reach due to Reach's fraud or Reach's Denylist Magento is displaying the error returned by Reach's Chekcout API directly to the customer which could tip-off fraudsters. To fix this, we will return a generic `PaymentAuthorizationFailed` error.

Reach Errors to be generic:
- Blacklisted 
- FraudSuspected

### DIFF
--- a/Model/Cc.php
+++ b/Model/Cc.php
@@ -693,7 +693,7 @@ class Cc extends \Magento\Payment\Model\Method\Cc
         if (isset($response['Error']) && count($response['Error'])) {
             $error = $response['Error']['Code'];
             if (isset($error) && $error != '') {
-                if ($error == "Blacklisted" || $error == "FraudSuspected") {
+                if ($error === "Blacklisted" || $error === "FraudSuspected") {
                     $errorMessage = ': PaymentAuthorizationFailed';
                 } else {
                     $errorMessage = ':'.$error;
--- a/Model/Cc.php
+++ b/Model/Cc.php
@@ -690,14 +690,17 @@ class Cc extends \Magento\Payment\Model\Method\Cc
      */
     public function processErrors($response)
     {
-
         if (isset($response['Error']) && count($response['Error'])) {
-            $errorMessage = $response['Error']['Code'];
-            if (isset($response['Error']['Message']) && $response['Error']['Message'] != '') {
-                $errorMessage = ':'.$response['Error']['Message'];
+            $error = $response['Error']['Code'];
+            if (isset($error) && $error != '') {
+                if ($error == "Blacklisted" || $error == "FraudSuspected") {
+                    $errorMessage = ': PaymentAuthorizationFailed';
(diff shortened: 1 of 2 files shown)


Model/Cc.php
```diff
@@ -690,14 +690,17 @@ protected function callCurl($url, $params, $method = "POST")
...
     public function processErrors($response)
     {
-
         if (isset($response['Error']) && count($response['Error'])) {
-            $errorMessage = $response['Error']['Code'];
-            if (isset($response['Error']['Message']) && $response['Error']['Message'] != '') {
-                $errorMessage = ':'.$response['Error']['Message'];
+            $error = $response['Error']['Code'];
+            if (isset($error) && $error != '') {
+                if ($error == "Blacklisted" || $error == "FraudSuspected") {
```
grandel-withreach: Could we switch the `==` to `===`?

Model/PayPalManagement.php
```diff
@@ -150,8 +150,14 @@ public function savePaymentAndPlaceOrder($cartId,$deviceFingerprint)
             }
         } catch (\Exception $e) {
             $this->response->setSuccess(false);
+            $error = $e->getMessage();
+            if ($error == "Blacklisted" || $error == "FraudSuspected") {
```
grandel-withreach: Could we switch the == to ===?

Model/Cc.php
```diff
@@ -690,14 +690,17 @@ protected function callCurl($url, $params, $method = "POST")
...
+                    $errorMessage = ': PaymentAuthorizationFailed';
+                } else {
+                    $errorMessage = ':'.$error;
```
grandel-withreach: Are the `':'` necessary? I think when we concat the string on line `703` we'll end up with double `:`

Model/PayPalManagement.php
```diff
@@ -150,8 +150,14 @@ public function savePaymentAndPlaceOrder($cartId,$deviceFingerprint)
...
+                $errorMessage = ': PaymentAuthorizationFailed';
```
grandel-withreach: Are the `':'` necessary? I think when we concat the string on line `160` we'll end up with double `:`

Model/Cc.php
ChantalDelachevrotiere: This was what was here previously with the other errors that I had seen.

grandel-withreach: Ohh I see, we then might want remove the colon in `Something went wrong while generating the payment request:` or not include the `:` in the values set for `$errorMessage`

ChantalDelachevrotiere: updated!

grandel-withreach (approved):
Looks good! thanks @ChantalDelachevrotiere

PR #92, opened 2022-07-28
Only return Saved Cards/Contracts tied to Customer Selected Currency

ChantalDelachevrotiere:
Contracts can only be used with the currency the contract was opened with. 

When presenting the saved cards (contracts) to the customer only saved cards (contracts) with the correct currency will be displayed.

### DIFF
--- a/Block/Customer/Savedcard.php
+++ b/Block/Customer/Savedcard.php
@@ -38,7 +38,7 @@ class Savedcard extends \Magento\Framework\View\Element\Template
         $customerCards = $this->getCustomerCards();
         if ($customerCards && $customerCards->count()) {
             foreach ($customerCards as $card) {
-                $cards[]=['type'=>$card->getMethod(),'number'=>$card->getIdentifier(),'contract_id'=>$card->getId()];
+                $cards[]=['type'=>$card->getMethod(),'number'=>$card->getIdentifier(),'contract_id'=>$card->getId(), 'currency'=>$card->getcurrency()];
             }
         }
         return $cards;
--- a/Block/Customer/Savedcard.php
+++ b/Block/Customer/Savedcard.php
@@ -38,7 +38,13 @@ class Savedcard extends \Magento\Framework\View\Element\Template
         $customerCards = $this->getCustomerCards();
         if ($customerCards && $customerCards->count()) {
             foreach ($customerCards as $card) {
-                $cards[]=['type'=>$card->getMethod(),'number'=>$card->getIdentifier(),'contract_id'=>$card->getId(), 'currency'=>$card->getcurrency()];
+                $cards[]=
+                    [
+                        'type'=>$card->getMethod(),
+                        'number'=>$card->getIdentifier(),
+                        'contract_id'=>$card->getId(),
+                        'currency'=>$card->getcurrency()
+                    ];
             }
         }
         return $cards;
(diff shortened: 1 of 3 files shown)


Block/Customer/Savedcard.php
```diff
@@ -38,7 +38,7 @@ public function getSavedCards()
         $customerCards = $this->getCustomerCards();
         if ($customerCards && $customerCards->count()) {
             foreach ($customerCards as $card) {
-                $cards[]=['type'=>$card->getMethod(),'number'=>$card->getIdentifier(),'contract_id'=>$card->getId()];
+                $cards[]=['type'=>$card->getMethod(),'number'=>$card->getIdentifier(),'contract_id'=>$card->getId(), 'currency'=>$card->getcurrency()];
```
grandel-withreach: Nit pick, but could we format this array to break each key-value pair into its own line

ex.

```
[
    'type' => $card->getMethod(),
    'number'=>$card->getIdentifier(),
    ...
]
```

Model/CcConfigProvider.php
```diff
@@ -106,7 +109,8 @@ public function getConfig()
                 'reach_cc' => [
                     'oc_enabled'=>(boolean)$this->reachHelper->getAllowOpenContract(),
                     'open_contracts'=>$this->getOpenContracts(),
-                     //got idea from here: https://webkul.com/blog/adding-additional-variables-in-window-checkoutconfig-on-magento-2-checkout-page/
+                    'selected_currency'=>$this->_storeManager->getStore()->getCurrentCurrency()->getCode(),
```
grandel-withreach, line 112: Oh nice! you were able to determine where the currency is being stored!

view/frontend/web/js/view/payment/method-renderer/cc.js
```diff
@@ -238,6 +240,7 @@ define([
                 {
                     return false;
                 }
+                console.log("herererere")
```
grandel-withreach: I presume this was unintentional 😂

Model/CcConfigProvider.php
```diff
@@ -134,7 +138,7 @@ protected function getOpenContracts()
             $collection = $this->openContract->getCollection();
             $collection->addFieldToFilter('customer_id', ['eq'=>$customerId]);
             foreach ($collection as $contract) {
-                $contracts[] = ['contractId'=>$contract->getReachContractId(),'label'=>$contract->getMethod().' - '.$contract->getIdentifier()];
+                $contracts[] = ['contractId'=>$contract->getReachContractId(),'label'=>$contract->getMethod().' - '.$contract->getIdentifier(), 'contractCurrency'=>$contract->getCurrency()];
```
grandel-withreach: Same formatting comment from above

Model/CcConfigProvider.php
ChantalDelachevrotiere: Well, sorta. At the time the contracts are being rendered on the FE the selected currency is not available. But this line of code solves that

view/frontend/web/js/view/payment/method-renderer/cc.js
ChantalDelachevrotiere: 💀

grandel-withreach (approved):

